PR #30, opened 2020-03-27
add Dockerfile

lispmeister:
This Dockerfile enables the build of a Docker image for the Aperture
binary.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM golang:1.13-alpine as builder
+
+# Copy in the local repository to build from.
+COPY . /go/src/github.com/lightninglabs/aperture
+
+# Force Go to use the cgo based DNS resolver. This is required to ensure DNS
+# queries required to connect to linked containers succeed.
+ENV GODEBUG netdns=cgo
+
+# Explicitly turn on the use of modules (until this becomes the default).
+ENV GO111MODULE on
+
+# Install dependencies and install/build aperture
+RUN apk add --no-cache --update alpine-sdk \
+    git \
+    make \
+&&  cd /go/src/github.com/lightninglabs/aperture/cmd \
+&&  go install ./...
+
+# Start a new, final image to reduce size.
+FROM alpine as final
+
+# Expose aperture ports
+EXPOSE 11010
+
+# Copy the binaries and entrypoint from the builder image.
+COPY --from=builder /go/bin/aperture /bin/
+
+# Add bash.
+RUN apk add --no-cache \
+    bash \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM golang:1.12-alpine as builder
```
guggero: Should be `go1.13` as per `go.mod`.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+COPY . /go/src/github.com/lightninglabs/aperture
+
+# Force Go to use the cgo based DNS resolver. This is required to ensure DNS
+# queries required to connect to linked containers succeed.
+ENV GODEBUG netdns=cgo
+
+# Explicitly turn on the use of modules (until this becomes the default).
+ENV GO111MODULE on
+
+# Install dependencies and install/build lnd.
```
guggero: nit: comment says `lnd`.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN apk add --no-cache --update alpine-sdk \
+    git \
+    make \
+&&  cd /go/src/github.com/lightninglabs/aperture/cmd \
+&&  go install ./...
+
+# Start a new, final image to reduce size.
+FROM alpine as final
+
+# Expose lnd ports (server, rpc).
```
guggero: nit: comment says `lnd`.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+EXPOSE 8081 11010
+
+# Copy the binaries and entrypoint from the builder image.
+COPY --from=builder /go/bin/aperture /bin/
+
+# Add bash.
+RUN apk add --no-cache \
+    bash \
+    ca-certificates
```
guggero: We should probably add an `ENTRYPOINT ["aperture"]` directive to the bottom.

Dockerfile
lispmeister: fixed

Dockerfile
lispmeister: fixed

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    git \
+    make \
+&&  cd /go/src/github.com/lightninglabs/aperture/cmd \
+&&  go install ./...
+
+# Start a new, final image to reduce size.
+FROM alpine as final
+
+# Expose aperture ports
+EXPOSE 8081
```
guggero, line 24: Port `11010` was actually correct, 8081 isn't used any more.

PR #35, opened 2020-04-28
Use same lnd's `cert` package to create TLS config to fix TLS cipher suites

guggero:
We need to use safe TLS1.2 cipher suites that are also accepted by the golang `http2` library. Turns out, that is a subset of what we currently use in `lnd`.


aperture.go
```diff
@@ -189,11 +189,11 @@ func setupLogging(cfg *config) error {
 // getTLSConfig returns a TLS configuration for either a self-signed certificate
 // or one obtained through Let's Encrypt.
 func getTLSConfig(serverName string, autoCert bool) (*tls.Config, error) {
-	var tlsKeyFile, tlsCertFile string
-	switch {
-	// When using autocert, we set a TLSConfig on the server so the key and
-	// cert file we pass in are ignored and don't need to exist.
-	case autoCert:
+	// If the --autocert flag is set, use the autocert library that will
```
joostjager, line 192: It feels a bit strange to mention the exact flag name here.

aperture.go
```diff
@@ -189,11 +189,11 @@ func setupLogging(cfg *config) error {
...
+	// create a new certificate for us from Let's Encrypt as soon as the
```
joostjager, line 193: "create ... from" ? request from?

aperture.go
```diff
@@ -231,6 +186,70 @@ func setupLogging(cfg *config) error {
...
+			}
+			log.Infof("Done generating TLS certificates")
+		}
+
+		certData, _, err := cert.LoadCert(tlsCertFile, tlsKeyFile)
+		if err != nil {
+			return nil, err
+		}
+
+		return cert.TLSConfFromCert(certData), nil
```
joostjager: This change is more than only a function extraction, probably deserves a mention in the commit message. Or a separate commit doing this first and the next commit extracting the function.

aperture.go
```diff
@@ -240,11 +240,44 @@ func getTLSConfig(serverName string, autoCert bool) (*tls.Config, error) {
...
 
-	certData, _, err := cert.LoadCert(tlsCertFile, tlsKeyFile)
+	certData, parsedCert, err := cert.LoadCert(tlsCertFile, tlsKeyFile)
 	if err != nil {
 		return nil, err
 	}
 
+	// If the certificate expired or it was outdated, delete it and the TLS
+	// key and generate a new pair.
+	if time.Now().After(parsedCert.NotAfter) {
```
joostjager, line 250: So the workflow is to wait until the certificates has expired and then restart the service? Sounds tricky to do without downtime.

Can't the certificate just never expire?

aperture.go
guggero: Agree.

aperture.go
guggero: Fixed the comment to be more precise.

aperture.go
guggero: This is the exact code that is used in `lnd`. But you are right, that doesn't make a lot of sense for a server that should have as little down time as possible.
No expiration is not possible, most TLS libraries will block a certificate that's valid for too long. But I chose a sane period that even the more restrictive libraries/OSes should allow and set the refresh check to about half that time. So as long as between year 1 and year 2 there is a restart, it should be fine.

guggero:
We ended up disabling TLS to get it to work. There seems to be a general problem with TLS and load balancers.
I decided to leave all the TLS changes in this PR anyway as this still might come in handy later on when we want to run aperture as a Kubernetes Ingress Controller.

aperture.go
```diff
@@ -34,6 +34,18 @@ const (
 	etcdKeyDelimeter = "/"
 )
 
+var (
+	// http2TLSCipherSuites is the list of cipher suites we allow the server
+	// to use. This list removes a CBC cipher from the list used in lnd's
+	// cert package because apparently HTTP/2 and recent go versions are
+	// more strict about what should be allowed.
```
joostjager, line 54: Is there a link to the http2 golang library that defines this?

aperture.go
```diff
@@ -138,6 +152,17 @@ func start() error {
 			}
 			log.Infof("Done generating TLS certificates")
 		}
+
+		// Load the certs now so we can create a complete TLS config.
+		certData, _, err := cert.LoadCert(tlsCertFile, tlsKeyFile)
```
joostjager: Cert loading needs to move to the next commit?

aperture.go
```diff
@@ -32,6 +32,19 @@ const (
...
+	// 14 months and is set to a maximum just below what some operating
+	// systems set as a sane maximum certificate duration. See
+	// https://support.apple.com/en-us/HT210176 for more information.
+	selfSignedCertValidity = time.Hour * 24 * 820
+
+	// selfSignedCertExpiryMargin is how much time before the certificate's
+	// expiry date we already refresh it with a new one. We set this to half
+	// the certificate validity length to make the chances bigger for it to
+	// be refreshed on a routine server restart.
+	selfSignedCertExpiryMargin = time.Hour * 24 * 410
```
joostjager, line 47: `selfSignedCertValidity / 2`?

aperture.go
```diff
@@ -247,19 +260,59 @@ func getTLSConfig(serverName string, autoCert bool) (*tls.Config, error) {
...
+	// Load the certs now so we can inspect it and return a complete TLS
+	// config later.
+	certData, parsedCert, err := cert.LoadCert(tlsCertFile, tlsKeyFile)
 	if err != nil {
 		return nil, err
 	}
+
+	// The margin is negative, so adding it to the expiry date should give
+	// us a date in about the middle of it's validity period.
+	expiryWithMargin := parsedCert.NotAfter.Add(
```
joostjager, line 292: `NotAfter.Sub`

proxy/proxy.go
```diff
@@ -70,8 +71,15 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	defer logRequest()
 
+	// PRI requests are an internal check of HTTP/2 which we just have to
+	// say OK to.
+	if r.Method == "PRI" {
```
joostjager, line 76: Some more info on this would be nice. What problem do we have if we don't answer?

proxy/proxy.go
```diff
@@ -148,8 +156,8 @@ func (p *Proxy) UpdateServices(services []*Service) error {
 	if err != nil {
 		return err
 	}
-	transport := &http.Transport{
-		ForceAttemptHTTP2: true,
+	transport := &http2.Transport{
+		AllowHTTP: true,
```
joostjager, line 160: Why this change?

proxy/proxy.go
```diff
@@ -315,6 +323,8 @@ func sendDirectResponse(w http.ResponseWriter, r *http.Request,
 	// Find out if the client is a normal HTTP or a gRPC client. Every gRPC
 	// request should have the Content-Type header field set accordingly
 	// so we can use that.
+	log.Debugf("Sending status code %d for request %s %s %s", statusCode,
+		r.Method, r.RequestURI, r.Proto)
```
joostjager, line 327: Does this need to stay?

aperture.go
```diff
@@ -113,9 +113,23 @@ func start() error {
 
 	// Create TLS configuration by either creating new self-signed certs or
 	// trying to obtain one through Let's Encrypt.
-	httpsServer.TLSConfig, err = getTLSConfig(cfg.ServerName, cfg.AutoCert)
-	if err != nil {
-		return err
+	var serveFn func() error
+	if cfg.Insecure {
+		serveFn = httpsServer.ListenAndServe
+		httpsServer.Handler = h2c.NewHandler(handler, &http2.Server{})
```
joostjager, line 119: Can also use some documentation about what `h2c` is

aperture.go
guggero: Not really, we can't create a valid `TLSConfig` without the certificate.

aperture.go
guggero: `Sub` takes a time, not a duration.

proxy/proxy.go
guggero: Removed the whole commit. We did try out some stuff with this but I think in the end (with TLS turned off, this is not needed anymore). Same with the following 2 commits.

guggero:
We tested this again on the AWS cluster and confirmed it still works.

aperture.go
joostjager: Ah that is right

aperture.go
```diff
@@ -114,9 +114,28 @@ func start() error {
 
 	// Create TLS configuration by either creating new self-signed certs or
 	// trying to obtain one through Let's Encrypt.
-	httpsServer.TLSConfig, err = getTLSConfig(cfg.ServerName, cfg.AutoCert)
-	if err != nil {
-		return err
+	var serveFn func() error
+	if cfg.Insecure {
+		// Normally, HTTP/2 only works with TLS. But there is a special
+		// version called HTTP/2 Cleartext (h2c) that some clients
```
joostjager, line 120: :+1:

Issue #7, opened 2019-08-22
custom price points per end point


Roasbeef:
The proxy should eventually grow its config to allow backend's to configure the precise value of the LSAT token for a particular endpoint. This value may even be dynamic, allowing an admin to increase/decrease the price via a programmatic interface. An example is the initial account creation for CLM.

ellemouton:
I put together a quick PoC that allows dynamic prices. I am not sure if it is completely the correct approach (but it works for my use case) so would love some feedback before cleaning it up and submitting a PR. 

What I did is add a `dynamicpriceendpoint` option to service config. If set, this means that aperture can expect there to be a gPRC server at that endpoint that serves price data given a resource/path. If the service is hit, then aperture queries the  gRPC server for prices. 

My PoC PR is [here](https://github.com/ellemouton/aperture/pull/1). To test this, I wrote a [simple example server](https://github.com/ellemouton/aperturetest) that implements the interface and returns different prices depending on the path (`book/1`, `book/2`, etc)

guggero:
Sounds interesting, thanks for the PoC! Will take a look tomorrow. Can you assign me as a reviewer on it please so I don't forget?

ellemouton:
awesome thanks! yeah i will do that 👍  i have sent an invite to you for the repo. Once accepted, I can assign you as a reviewer 😊

guggero:
Great, thanks! Was even able to assign myself now.

guggero:
Fixed by #36.

PR #36, opened 2020-05-10
aperture: Custom price per service

ellemouton:
This PR addresses part of #7 . It allows you to add a static price for each service in your config as at the moment the value of any LSAT you mint is 1 satoshi.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -96,10 +96,10 @@ func start() error {
 	}
 
 	// Create the proxy and connect it to lnd.
-	genInvoiceReq := func() (*lnrpc.Invoice, error) {
+	genInvoiceReq := func(price int64) (*lnrpc.Invoice, error) {
 		return &lnrpc.Invoice{
 			Memo:  "LSAT",
-			Value: 1,
+			Value: price,
 		}, nil
 	}
 	servicesProxy, err := createProxy(cfg, genInvoiceReq, etcdClient)
--- a/auth/authenticator.go
+++ b/auth/authenticator.go
@@ -59,9 +59,13 @@ func (l *LsatAuthenticator) Accept(header *http.Header, serviceName string) bool
 //
 // NOTE: This is part of the Authenticator interface.
 func (l *LsatAuthenticator) FreshChallengeHeader(r *http.Request,
-	serviceName string) (http.Header, error) {
+	serviceName string, servicePrice int64) (http.Header, error) {
 
-	service := lsat.Service{Name: serviceName, Tier: lsat.BaseTier}
+	service := lsat.Service{
+		Name:  serviceName,
+		Tier:  lsat.BaseTier,
+		Price: servicePrice,
+	}
 	mac, paymentRequest, err := l.minter.MintLSAT(
 		context.Background(), service,
 	)
--- a/lsat/service.go
+++ b/lsat/service.go
@@ -46,6 +46,9 @@ type Service struct {
 
 	// Tier is the tier of the LSAT-enabled service.
 	Tier ServiceTier
+
+	// Price of service LSAT in satoshis.
+	Price int64
 }
 
 // NewServicesCaveat creates a new services caveat with the provided caveats.
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -97,7 +97,7 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case authLevel.IsOn():
 		if !p.authenticator.Accept(&r.Header, target.Name) {
 			prefixLog.Infof("Authentication failed. Sending 402.")
-			p.handlePaymentRequired(w, r, target.Name)
+			p.handlePaymentRequired(w, r, target.Name, target.Price)
 			return
 		}
 
@@ -116,7 +116,7 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 			if !ok {
-				p.handlePaymentRequired(w, r, target.Name)
+				p.handlePaymentRequired(w, r, target.Name, target.Price)
 				return
 			}
 			_, err = target.freebieDb.TallyFreebie(r, remoteIP)
@@ -281,11 +281,11 @@ func addCorsHeaders(header http.Header) {
 // handlePaymentRequired returns fresh challenge header fields and status code
 // to the client signaling that a payment is required to fulfil the request.
 func (p *Proxy) handlePaymentRequired(w http.ResponseWriter, r *http.Request,
-	serviceName string) {
+	serviceName string, servicePrice int64) {
 
 	addCorsHeaders(r.Header)
 
-	header, err := p.authenticator.FreshChallengeHeader(r, serviceName)
+	header, err := p.authenticator.FreshChallengeHeader(r, serviceName, servicePrice)
 	if err != nil {
 		log.Errorf("Error creating new challenge header: %v", err)
 		sendDirectResponse(
--- a/proxy/service.go
+++ b/proxy/service.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/btcsuite/btcutil"
 	"github.com/lightninglabs/aperture/auth"
 	"github.com/lightninglabs/aperture/freebie"
 )
@@ -17,6 +18,16 @@ var (
 	filePrefix       = "!file"
 	filePrefixHex    = filePrefix + "+hex"
 	filePrefixBase64 = filePrefix + "+base64"
+)
+
+const (
+	// defaultServicePrice is price in satoshis to be used as the default
+	// service price.
+	defaultServicePrice = 1
+
+	// maxServicePrice is the maximum price in satoshis that can be used
+	// to create an invoice through lnd.
+	maxServicePrice = btcutil.SatoshiPerBitcoin * 100000
 )
 
 // Service generically specifies configuration data for backend services to the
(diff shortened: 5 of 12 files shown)


aperture.go
```diff
@@ -209,6 +209,13 @@ func getConfig(configFile string) (*config, error) {
 	if cfg.ListenAddr == "" {
 		return nil, fmt.Errorf("missing listen address for server")
 	}
+
+	for _, service := range cfg.Services {
+		if service.Price == 0 {
```
guggero, line 214: I think we should define a default value of 1 satoshi and use that if no price is defined on a service. See comment further below. But we should check if the price is negative and then error out (because the field is defined as `int64`).

auth/authenticator.go
```diff
@@ -59,9 +59,9 @@ func (l *LsatAuthenticator) Accept(header *http.Header, serviceName string) bool
 //
 // NOTE: This is part of the Authenticator interface.
 func (l *LsatAuthenticator) FreshChallengeHeader(r *http.Request,
-	serviceName string) (http.Header, error) {
+	serviceName string, servicePrice int64) (http.Header, error) {
 
-	service := lsat.Service{Name: serviceName, Tier: lsat.BaseTier}
+	service := lsat.Service{Name: serviceName, Tier: lsat.BaseTier, Price: servicePrice}
```
guggero, line 64: nit: line too long.

lsat/service.go
```diff
@@ -46,6 +46,9 @@ type Service struct {
 
 	// Tier is the tier of the LSAT-enabled service.
 	Tier ServiceTier
+
+	// Price of service LSAT in satoshis
```
guggero, line 50: nit: missing full stop.

proxy/service.go
```diff
@@ -69,6 +69,10 @@ type Service struct {
 	// correspond to the caveat's condition.
 	Constraints map[string]string `long:"constraints" description:"The service constraints to enforce at the base tier"`
 
+	// StaticPrice is the custom LSAT value in satoshis to be used for the
```
guggero, line 72: nit: Godoc comment doesn't match variable name.

proxy/proxy.go
```diff
@@ -281,11 +281,11 @@ func addCorsHeaders(header http.Header) {
 // handlePaymentRequired returns fresh challenge header fields and status code
 // to the client signaling that a payment is required to fulfil the request.
 func (p *Proxy) handlePaymentRequired(w http.ResponseWriter, r *http.Request,
-	serviceName string) {
+	serviceName string, servicePrice int64) {
 
 	addCorsHeaders(r.Header)
 
-	header, err := p.authenticator.FreshChallengeHeader(r, serviceName)
+	header, err := p.authenticator.FreshChallengeHeader(r, serviceName, servicePrice)
```
guggero, line 288: We could check if the price is `0` here and use the default price instead.

ellemouton:
Thanks for the comments @guggero! I have updated the PR as suggested 👍

auth/authenticator.go
```diff
@@ -59,9 +59,10 @@ func (l *LsatAuthenticator) Accept(header *http.Header, serviceName string) bool
 //
 // NOTE: This is part of the Authenticator interface.
 func (l *LsatAuthenticator) FreshChallengeHeader(r *http.Request,
-	serviceName string) (http.Header, error) {
+	serviceName string, servicePrice int64) (http.Header, error) {
 
-	service := lsat.Service{Name: serviceName, Tier: lsat.BaseTier}
+	service := lsat.Service{Name: serviceName, Tier: lsat.BaseTier,
```
guggero, line 64: nit: if the struct initialization doesn't fit on one line, the preferred style is this one:
```
service := lsat.Service{
	Name: serviceName,
	Tier: lsat.BaseTier,
	Price: servicePrice,
}
```

proxy/proxy.go
```diff
@@ -281,11 +281,17 @@ func addCorsHeaders(header http.Header) {
...
-	serviceName string) {
+	serviceName string, servicePrice int64) {
 
 	addCorsHeaders(r.Header)
 
-	header, err := p.authenticator.FreshChallengeHeader(r, serviceName)
+	// Use the default price of 1 satoshi if servicePrice is zero
+	price := servicePrice
+	if servicePrice == 0 {
+		price = 1
```
guggero, line 291: nit: would be nice to define this as a constant.

services.go
```diff
@@ -27,7 +27,7 @@ func newStaticServiceLimiter(proxyServices []*proxy.Service) *staticServiceLimit
 	constraints := make(map[lsat.Service][]lsat.Caveat)
 
 	for _, proxyService := range proxyServices {
-		s := lsat.Service{Name: proxyService.Name, Tier: lsat.BaseTier}
+		s := lsat.Service{Name: proxyService.Name, Tier: lsat.BaseTier, Price: proxyService.Price}
```
guggero, line 30: nit: line too long.

proxy/proxy.go
ellemouton: Cool yeah, good idea 👍 Where would be a good place to define this constant?

ellemouton: Seems weird to have the constant defined in proxy.go? for now i will define it in proxy/service.go. But let me know if I should move it 👍

guggero: I think for now just at the top of the same file.

ellemouton: Ok cool 👍

proxy/proxy.go
```diff
@@ -281,11 +282,17 @@ func addCorsHeaders(header http.Header) {
...
-	serviceName string) {
+	serviceName string, servicePrice int64) {
 
 	addCorsHeaders(r.Header)
 
-	header, err := p.authenticator.FreshChallengeHeader(r, serviceName)
+	// Use the default price of 1 satoshi if servicePrice is zero
+	price := servicePrice
+	if servicePrice == 0 {
+		price = defaultServicePrice
```
wpaulino, line 292: This would be better to set when parsing the config so that we don't have to do this every time elsewhere.

proxy/proxy.go
```diff
@@ -22,9 +22,10 @@ const (
 	// An example entry would look like this:
 	// 2019-11-09 04:07:55.072 [INF] PRXY: 66.249.69.89 - -
 	// "GET /availability/v1/btc.json HTTP/1.1" "" "Mozilla/5.0 ..."
-	formatPattern  = "- - \"%s %s %s\" \"%s\" \"%s\""
-	hdrContentType = "Content-Type"
-	hdrTypeGrpc    = "application/grpc"
+	formatPattern       = "- - \"%s %s %s\" \"%s\" \"%s\""
+	hdrContentType      = "Content-Type"
+	hdrTypeGrpc         = "application/grpc"
+	defaultServicePrice = 1
```
wpaulino, line 28: This can then be moved elsewhere, either `aperture.go` where we parse the config, or in `proxy/service.go`.

aperture.go
```diff
@@ -209,6 +209,14 @@ func getConfig(configFile string) (*config, error) {
 	if cfg.ListenAddr == "" {
 		return nil, fmt.Errorf("missing listen address for server")
 	}
+
+	for _, service := range cfg.Services {
+		if service.Price < 0 {
+			return nil, fmt.Errorf("negative price set for "+
+				"service %s", service.Name)
+		}
```
wpaulino, line 217: Could you also add a check for the max? `lnd` prevents creating invoices above 100,000 BTC.

proxy/proxy.go
wpaulino: Could you also add a debug log along the lines of `Using default LSAT price of %v per service`?

proxy/proxy.go
wpaulino: Could you also add a comment noting the units of the constant?

aperture.go
ellemouton: Is it fine if i use 'lnwire.MaxMilliSatoshi.ToSatoshis() as a max?

ellemouton:
Thanks for the comments @wpaulino 😄 I have updated accordingly.

aperture.go
wpaulino: It can still fail since `lnd` enforces the 100,000 BTC maximum at the RPC level, which is less than `lnwire.MaxMilliSatoshi.ToSatoshis()`.

aperture.go
```diff
@@ -209,6 +214,22 @@ func getConfig(configFile string) (*config, error) {
 	if cfg.ListenAddr == "" {
 		return nil, fmt.Errorf("missing listen address for server")
 	}
+
+	for _, service := range cfg.Services {
+		switch {
+		case service.Price == 0:
+			log.Infof("Using default LSAT price of %v satoshis for "+
```
wpaulino, line 221:
```suggestion
			log.Debugf("Using default LSAT price of %v satoshis for "+
```

aperture.go
```diff
@@ -209,6 +218,22 @@ func getConfig(configFile string) (*config, error) {
 	if cfg.ListenAddr == "" {
 		return nil, fmt.Errorf("missing listen address for server")
 	}
+
+	for _, service := range cfg.Services {
+		switch {
```
guggero, line 223: Nice to have all the checks in one place now! Looking at it again, I think this `switch` (and the two related constants) should go into the loop of `prepareServices` in `proxy/service.go`. Sorry to request yet another code move... But the thing is, the log statement also doesn't work here because the log subsystem hasn't been initialized at this point.

ellemouton: cool, good idea 👍

ellemouton:
Updated as per suggestion @guggero 🚀

WalterSmuts:
Such a cool feature! :P

PR #39, opened 2020-07-24
auth: check LSAT invoice status by payment hash

guggero:
To make sure an invoice of an LSAT is actually paid, we check the invoice status in the backing `lnd` node in addition to the preimage.


test/chainnotifier_mock.go
```diff
@@ -0,0 +1,160 @@
+package test
```
wpaulino, line 1: Perhaps this code should've also been moved to `lndclient` so that we don't have to copy it everywhere?

challenger.go
```diff
@@ -50,13 +69,129 @@ func NewLndChallenger(cfg *authConfig, genInvoiceReq InvoiceRequestGenerator) (
...
+			default:
+			}
+
+			// Wait for an update to arrive. This will block until
+			// either a message receives, an error occurs or the
+			// underlying context is canceled (which will also
+			// result in an error).
+			invoice, err := subscriptionResp.Recv()
+			switch {
+			case err != nil && strings.Contains(
```
wpaulino, line 172: Check EOF here too?

challenger.go
```diff
@@ -79,3 +214,39 @@ func (l *LndChallenger) NewChallenge(price int64) (string, lntypes.Hash, error)
...
+		l.invoicesMtx.Unlock()
+
+		// Exit early if we have the invoice with the correct status.
+		if ok && invoiceState == state {
+			log.Debugf("Invoice for hash %v has status: %v",
+				hash, invoiceState)
+			return nil
+		}
+
+		<-time.After(invoicePollInterval)
```
wpaulino, line 258: Should also check quit here.

challenger.go
```diff
@@ -79,3 +214,39 @@ func (l *LndChallenger) NewChallenge(price int64) (string, lntypes.Hash, error)
...
+
+		select {
+		case <-exitTimer:
```
wpaulino, line 261: And here as well.

auth/authenticator.go
```diff
@@ -51,6 +58,16 @@ func (l *LsatAuthenticator) Accept(header *http.Header, serviceName string) bool
 		return false
 	}
 
+	// Make sure the backend has the invoice recorded as settled.
+	err = l.checker.VerifyInvoiceStatus(
```
wpaulino, line 62: Just to confirm, will the timeout here block processing further requests? I believe `ServeHTTP` is spawned as a goroutine so it should be fine?

challenger_test.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return &LndChallenger{
+		client:        mockClient,
+		genInvoiceReq: genInvoiceReq,
+		invoiceStates: make(map[lntypes.Hash]lnrpc.Invoice_InvoiceState),
+		quit:          make(chan struct{}),
+	}, mockClient
+}
+
+func TestLndChallenger(t *testing.T) {
```
wpaulino, line 105: Nit: commit message isn't fully descriptive of the changes, namely the added test.

test/signer_mock.go
```diff
@@ -0,0 +1,58 @@
+package test
```
Roasbeef, line 1: Not a blocker, but we could make this an internal package to signal to users of the project that it shouldn't be used externally (an internal package will actually prevent it from being imported in another project). Not sure if that's what we want here though.

auth/authenticator.go
Roasbeef: Yeah IIRC it is indeed a distinct goroutine.

auth/interface.go
```diff
@@ -30,3 +39,14 @@ type Minter interface {
 	// VerifyLSAT attempts to verify an LSAT with the given parameters.
 	VerifyLSAT(context.Context, *mint.VerificationParams) error
 }
+
+// InvoiceChecker is an entity that is able to check the status of an invoice,
+// particularly whether it's been paid or not.
+type InvoiceChecker interface {
+	// VerifyInvoiceStatus checks that an invoice identified by a payment
```
Roasbeef, line 46: Given the control flow above, a user will only enter this verification phase once they have the pre-image, correct? So at this point there shouldn't be any risk of the user gaining the pre-image before we settle things on our end.

challenger.go
```diff
@@ -19,14 +30,22 @@ type InvoiceRequestGenerator func(price int64) (*lnrpc.Invoice, error)
 type LndChallenger struct {
 	client        lnrpc.LightningClient
 	genInvoiceReq InvoiceRequestGenerator
+
+	invoiceStates  map[lntypes.Hash]lnrpc.Invoice_InvoiceState
```
Roasbeef, line 52: Hmm, I guess it would take a _ton_ of invoices for this to really become an issue memory wise. In the future though, we should definitely limit the number of entries that we'll accumulate here.

challenger.go
```diff
@@ -50,13 +69,129 @@ func NewLndChallenger(cfg *authConfig, genInvoiceReq InvoiceRequestGenerator) (
...
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	// Advance our indices to the latest known one so we'll only receive
+	// updates for new invoices and/or newly settled invoices.
+	l.invoicesMtx.Lock()
+	for _, invoice := range invoiceResp.Invoices {
```
Roasbeef, line 123: I think it should be safe to ignore known _expired_ invoices here. If we want to cut things down even further, we can use the current LSAT price as well (ignore invoices that don't match value-wise).

challenger.go
```diff
@@ -50,13 +69,129 @@ func NewLndChallenger(cfg *authConfig, genInvoiceReq InvoiceRequestGenerator) (
...
+		cancel()
+		return err
+	}
+
+	l.wg.Add(1)
+	go func() {
+		defer l.wg.Done()
+
+		for {
+			// In case we receive the shutdown signal right after
```
Roasbeef, line 158: `defer cancel()`?

challenger.go
```diff
@@ -79,3 +214,39 @@ func (l *LndChallenger) NewChallenge(price int64) (string, lntypes.Hash, error)
...
+// invoice update is still on its way, we try several times until either
+// the desired status is set or the given timeout is reached.
+//
+// NOTE: This is part of the auth.InvoiceChecker interface.
+func (l *LndChallenger) VerifyInvoiceStatus(hash lntypes.Hash,
+	state lnrpc.Invoice_InvoiceState, timeout time.Duration) error {
+
+	exitTimer := time.After(timeout)
+	for {
+		l.invoicesMtx.Lock()
```
Roasbeef, line 247: If we want to avoid spinning here (even with the interval) we can use another map to store a series of [condition variables](https://golang.org/pkg/sync/#Cond), which will be signaled once the entry is populated. A rough sketch would be something like:
  * grab lock
  * check if invoice is there
  * if so, gg
  * if not, add a new condition variable to the signal map (specific for this invoice) 
  * wait until signaled 

Alternatively, we can just use a single condition variable for (invoice state has been updated), which'll let us avoid needing to tack on yet another map. This route is simpler and lets us avoid polling the map, making everything a bit more async.

challenger.go
Roasbeef: If we go with what I described above, we can eliminate the timer all together.

aperture.go
```diff
@@ -179,17 +187,51 @@ func start() error {
...
+			select {
+			case errChan <- torHTTPServer.ListenAndServe():
+			case <-quit:
+			}
 		}()
-		defer httpServer.Close()
 	}
 
-	return <-errChan
+	// Now that we've started everything, intercept any interrupt signals
```
Roasbeef, line 205: 👍

test/chainnotifier_mock.go
guggero: There was quite some discussion around this in loop. The server also used that test code and imported it from the client repo. But then the server required the mocks to be more specific and have more features, so every time someone wanted to extend the tests in the server repo a change in the client repo was also necessary. That's why it was instead copied to the server and no longer imported.

Moving the code to `lndclient` doesn't really solve the problem of having different requirements for the mocks in different projects.

auth/authenticator.go
guggero: Affirmative: https://golang.org/src/net/http/server.go#L2962

auth/interface.go
guggero: The whole timeout thing was added to make sure the call from the client (that includes the preimage) doesn't arrive before we get the update from our invoices subscription gRPC stream, in case `lnd` is slow to send it. Or what exactly do you mean?

challenger.go
guggero: Very nice idea! I've never used the `sync.Cond` before, so I went with the suggestion. The code _looks_ a bit more complicated but I think it's not too bad. And there certainly isn't any spinning involved anymore.

challenger.go
```diff
@@ -47,16 +62,156 @@ func NewLndChallenger(cfg *authConfig, genInvoiceReq InvoiceRequestGenerator) (
...
+				log.Errorf("Error parsing invoice hash: %v",
+					err)
+				return
+			}
+
+			l.invoicesMtx.Lock()
+			if invoiceIrrelevant(invoice) {
+				// Don't keep the state of canceled or expired
+				// invoices.
+				delete(l.invoiceStates, hash)
```
Roasbeef, line 207: 👍

challenger.go
```diff
@@ -47,16 +62,156 @@ func NewLndChallenger(cfg *authConfig, genInvoiceReq InvoiceRequestGenerator) (
...
+			} else {
+				l.invoiceStates[hash] = invoice.State
+			}
+
+			// Before releasing the lock, notify our conditions that
+			// listen for updates on the invoice state.
+			l.invoicesCond.Broadcast()
```
Roasbeef, line 214: 👍

challenger.go
```diff
@@ -79,3 +234,104 @@ func (l *LndChallenger) NewChallenge(price int64) (string, lntypes.Hash, error)
...
+	go func() {
+		defer condWg.Done()
+		l.invoicesCond.L.Lock()
+
+		// Block here until our condition is met or the allowed time is
+		// up. The Wait() will return whenever a signal is broadcast.
+		invoiceState, hasInvoice = l.invoiceStates[hash]
+		for !(hasInvoice && invoiceState == state) && !timeoutReached {
+			l.invoicesCond.Wait()
+
```
Roasbeef, line 310: Check the quit signal here to make sure the goroutine is able to exit when we need to do an early shutdown, but the invoice isn't yet known/settled?

guggero: The goroutine above would exit on `l.quit` and set `timeoutReached` to `true` and also broadcast an update on the `Cond` object. That would cause the `Wait()` to return and we wouldn't enter the loop again afterward because `timeoutReached` is now true. Or what am I missing?

challenger.go
```diff
@@ -47,16 +80,156 @@ func NewLndChallenger(cfg *authConfig, genInvoiceReq InvoiceRequestGenerator) (
...
+	for _, invoice := range invoiceResp.Invoices {
+		if invoice.AddIndex > addIndex {
+			addIndex = invoice.AddIndex
+		}
+		if invoice.SettleIndex > settleIndex {
+			settleIndex = invoice.SettleIndex
+		}
+		hash, err := lntypes.MakeHash(invoice.RHash)
+		if err != nil {
+			return fmt.Errorf("error parsing invoice hash: %v", err)
```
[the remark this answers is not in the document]
guggero: Good catch, fixed. Probably not a problem as you say but nice to be correct anyway.

internal/chainnotifier_mock.go
```diff
@@ -0,0 +1,160 @@
+package test
```
wpaulino, line 1: Nit: typically `internal` is just a top-level directory containing subdirectories which are the internal packages.

challenger.go
```diff
@@ -47,16 +62,157 @@ func NewLndChallenger(cfg *authConfig, genInvoiceReq InvoiceRequestGenerator) (
...
+			SettleIndex: settleIndex,
+		},
+	)
+	if err != nil {
+		cancel()
+		return err
+	}
+
+	l.wg.Add(1)
+	go func() {
```
wpaulino, line 160: Nit: extract into method?

internal/chainnotifier_mock.go
guggero: Ah, gotcha. Fixed.

PR #40, opened 2020-08-14
aperture: only renew certificate if we signed it

guggero:
In case we use an externally provided certificate (which is created by
an external Let's Encrypt process for example), we don't want to try to
renew it. We identify our own certificates by the organization field we
set.


Roasbeef:
Will confirm that this fixes our issue before merging.

PR #41, opened 2020-08-22
aperture: relax TLS requirements

Roasbeef:
In this commit, we modify our cipher suites and required TLS versions to
allow anything greater than TL 1.1. TLS 1.0 (sslv3) is broken so we
require versions that're safely above that. Without this change, widely
used clients such as `openssl` will fail to connect out to an Aperture
proxy.


guggero:
tACK :+1: 
Tested locally on regtest with gRPC (loop and node.js) and HTTP/2 in Firefox (against demo HTML page).
For node.js the `process.env.GRPC_SSL_CIPHER_SUITES` variable now doesn't need to be set anymore. It works with an empty env var or with the value `HIGH+ECDSA`.

guggero (approved):

PR #46, opened 2020-09-23
challenger: shutdown if connection to lnd is lost

guggero:
Fixes the issue `Deny: Invoice status mismatch: no active or settled invoice found for hash=5e0...` when aperture loses its connection to lnd. For now we just shut down so an orchestrator can restart it when the connection is available again.

### DIFF
--- a/challenger.go
+++ b/challenger.go
@@ -191,17 +191,39 @@ func (l *LndChallenger) readInvoiceStream(
 		switch {
 
 		case err == io.EOF:
+			// The connection is shutting down, we can't continue
+			// to function properly. Signal the error to the main
+			// goroutine to force a shutdown/restart.
+			select {
+			case l.errChan <- err:
+			case <-l.quit:
+			default:
+			}
+
 			return
 
 		case err != nil && strings.Contains(
 			err.Error(), context.Canceled.Error(),
 		):
 
+			// The context has been canceled, we are shutting down.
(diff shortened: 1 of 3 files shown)


challenger.go
```diff
@@ -191,17 +191,39 @@ func (l *LndChallenger) readInvoiceStream(
...
 
 		case err != nil:
 			log.Errorf("Received error from invoice subscription: "+
 				"%v", err)
+
+			// The connection is faulty, we can't continue to
+			// function properly. Signal the error to the main
+			// goroutine to force a shutdown/restart.
+			select {
+			case l.errChan <- err:
```
wpaulino, line 222: Would be nice if we could determine whether this error is connection related so that we could just continue operation otherwise, but there doesn't seem to be a good way to do so.

wpaulino (commented):
Have we determined how often this happens? As it'll also affect the uptime of the services gated behind the proxy.

guggero:
It seems this only happens when lnd is re-deployed or restarted. Shutting down aperture and letting it be restarted should be a matter of a second or two. If we don't restart aperture, no new LSATs can be verified which is definitely worse for the uptime of the services gated behind the proxy.

wpaulino (approved):

Roasbeef (approved):
LGTM 🎷

Solid find! Makes sense why we saw this at times when a pod was restarted or something.

PR #48, opened 2020-12-02
maintenance: update lnd dependencies, prepare for production docker images

guggero:
We update to `lnd v0.11.1-beta` and `lndclient v0.11.0-4` and add a new production `Dockerfile` and a GitHub workflow that is triggered whenever a new version tag is
pushed. It will trigger a docker image build for that version and automatically push it to the specified repo.

For this to work, the GitHub repository needs the following two secrets to be configured:

```
DOCKER_USERNAME: <the user to log into Docker Hub>
DOCKER_API_KEY: <an API key generated by Docker Hub for the above user>
```

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,44 @@
+name: Docker image build
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  DOCKER_REPO: lightninglabs
+  DOCKER_IMAGE: aperture
+
+jobs:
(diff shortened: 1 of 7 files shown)


wpaulino (approved):
Is there a way to confirm whether the workflow would succeed without merging?

guggero:
Yes, you can add that commit to your fork's master branch, add the docker hub API secrets to your fork's GH settings and then push a tag. I did this for `lnd` and it's basically the exact same script with just a different repo name.

Roasbeef (approved):
LGTM 🥨

Secrets in place!

PR #49, opened 2020-12-29
Kirin comments renamed to Aperture

fiksn:
When starting to study how the code works I've just stumbled across a few instances of Kirin (which according to 0c14706fa4e8c4903e0b7d6ba79de3a5324c65b9 is the old name). Trivial search & replace (with sed) to move everything to Aperture. Hopefully after I get a better understanding of this I can contribute something more useful too.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -67,7 +67,7 @@ var (
 	}
 )
 
-// Main is the true entrypoint of Kirin.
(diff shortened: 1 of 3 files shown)


guggero (approved):
LGTM, thanks!

PR #50, opened 2020-12-31
Be able to run aperture without etcd

fiksn:
As promised in https://github.com/lightninglabs/aperture/pull/49 my wish is to contribute something more useful here too. I understand this will probably need more discussion, just to see whether my idea is sane. Golang is not my native language, so I am of course also open to suggestions in that direction.

Etcd is quite useful and lightweight, but I want to be able to run aperture w/o that dependency.

This is done through:
- add simple file-backed storage for onion keys
- be able to use pseudo-random secrets for LSAT macaroons
(unpredictable seed from config file and random TokenID is used for
that in a HMAC construction. Macaroons are HMACs all the way down
anyways ;) )

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -2,12 +2,14 @@ package aperture
 
 import (
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
@@ -21,6 +23,7 @@ import (
 	"github.com/lightningnetwork/lnd/signal"
 	"github.com/lightningnetwork/lnd/tor"
 	"golang.org/x/crypto/acme/autocert"
+	"golang.org/x/crypto/scrypt"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
 	"gopkg.in/yaml.v2"
@@ -51,6 +54,12 @@ const (
 	// the certificate validity length to make the chances bigger for it to
 	// be refreshed on a routine server restart.
 	selfSignedCertExpiryMargin = selfSignedCertValidity / 2
+
+	// minUserSeedLength is how long user specified seed should at least be
+	minUserSeedLength = 16
+
+	// keySize is how long the key derived from user provided seed will be
+	keySize = 32
 )
 
 var (
@@ -98,15 +107,19 @@ func run() error {
 		return err
 	}
 
+	var etcdClient *clientv3.Client
 	// Initialize our etcd client.
-	etcdClient, err := clientv3.New(clientv3.Config{
-		Endpoints:   []string{cfg.Etcd.Host},
-		DialTimeout: 5 * time.Second,
-		Username:    cfg.Etcd.User,
-		Password:    cfg.Etcd.Password,
-	})
-	if err != nil {
-		return fmt.Errorf("unable to connect to etcd: %v", err)
+	if cfg.Etcd != nil && cfg.Etcd.Host != "" {
+		etcdClient, err = clientv3.New(clientv3.Config{
+			Endpoints:   []string{cfg.Etcd.Host},
+			DialTimeout: 5 * time.Second,
+			Username:    cfg.Etcd.User,
+			Password:    cfg.Etcd.Password,
+		})
+		if err != nil {
+			log.Warnf("unable to connect to etcd: %v", err)
+			etcdClient = nil
+		}
 	}
 
 	// Create our challenger that uses our backing lnd node to create
@@ -193,6 +206,7 @@ func run() error {
 	// relatively safe.
 	var torHTTPServer *http.Server
 	if cfg.Tor != nil && (cfg.Tor.V2 || cfg.Tor.V3) {
+
 		torController, err := initTorListener(cfg, etcdClient)
 		if err != nil {
 			return err
@@ -415,10 +429,21 @@ func getTLSConfig(serverName string, autoCert bool) (*tls.Config, error) {
 func initTorListener(cfg *config, etcd *clientv3.Client) (*tor.Controller, error) {
 	// Establish a controller connection with the backing Tor server and
 	// proceed to create the requested onion services.
+
+	var store tor.OnionStore
+	if cfg.Tor.FileStorage {
+		store = newOnionStoreFile(apertureDataDir)
+	} else {
+		if etcd == nil {
+			return nil, errors.New("tor requires etcd")
+		}
+		store = newOnionStore(etcd)
+	}
+
 	onionCfg := tor.AddOnionConfig{
 		VirtualPort: int(cfg.Tor.VirtualPort),
 		TargetPorts: []int{int(cfg.Tor.ListenPort)},
-		Store:       newOnionStore(etcd),
+		Store:       store,
 	}
 	torController := tor.NewController(cfg.Tor.Control, "", "")
 	if err := torController.Start(); err != nil {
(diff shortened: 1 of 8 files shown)


aperture.go
```diff
@@ -98,15 +98,19 @@ func run() error {
...
 	// Initialize our etcd client.
-	etcdClient, err := clientv3.New(clientv3.Config{
-		Endpoints:   []string{cfg.Etcd.Host},
-		DialTimeout: 5 * time.Second,
-		Username:    cfg.Etcd.User,
-		Password:    cfg.Etcd.Password,
-	})
-	if err != nil {
-		return fmt.Errorf("unable to connect to etcd: %v", err)
+	if cfg.Etcd != nil {
```
guggero, line 103: This might be non-nil if someone just comments out the sub keys. Perhaps we need a new top-level option, for example `backend-mode` that is either `etcd` or `lightweight`. That would replace the `staticSecret.Enabled` and `tor.Filestorage` flags.

aperture.go
```diff
@@ -415,10 +420,21 @@ func getTLSConfig(serverName string, autoCert bool) (*tls.Config, error) {
 func initTorListener(cfg *config, etcd *clientv3.Client) (*tor.Controller, error) {
 	// Establish a controller connection with the backing Tor server and
 	// proceed to create the requested onion services.
+
+	var store tor.OnionStore
+	if cfg.Tor.FileStorage {
+		store = newOnionStoreFile(apertureDataDir)
+	} else {
+		if etcd == nil {
+			log.Errorf("Tor requires etcd")
```
guggero, line 429: Non-recoverable errors should be returned instead of just logged.

aperture.go
```diff
@@ -452,10 +468,20 @@ func initTorListener(cfg *config, etcd *clientv3.Client) (*tor.Controller, error
 func createProxy(cfg *config, challenger *LndChallenger,
 	etcdClient *clientv3.Client) (*proxy.Proxy, error) {
 
+	var seed *string
```
guggero, line 471: `nil` strings are used rarely in golang. It's totally fine to just use `var seed string` and later do a `if seed == ""` comparison.

mint/mint.go
```diff
@@ -122,7 +140,7 @@ func (m *Mint) MintLSAT(ctx context.Context,
 	)
 	if err != nil {
 		// Attempt to revoke the secret to save space.
-		_ = m.cfg.Secrets.RevokeSecret(ctx, idHash)
+		_ = cleanup(ctx)
```
guggero, line 143: If the error is never checked, it can just be removed from the returned function.

mint/mint.go
```diff
@@ -231,6 +249,55 @@ type VerificationParams struct {
...
+	if m.cfg.SeedForPseudoRandomness == nil {
+		return m.cfg.Secrets.GetSecret(
+			ctx, sha256.Sum256(params.Macaroon.Id()),
+		)
+	}
+	return m.generatePseudoRandomSecret(id)
+}
+
+// Creates a new secret (persistently stored or pseudo-random).
+func (m *Mint) createSecret(ctx context.Context, secretID [32]byte, id *lsat.Identifier) ([lsat.SecretSize]byte, SecretRemover, error) {
```
guggero, line 283: nit: We try to adhere to `lnd`'s [code guidelines](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md) and style. For example, everything should be wrapped before column 80 (where possible), with the tab character counting as 8.

aperture.go
```diff
@@ -452,10 +468,20 @@ func initTorListener(cfg *config, etcd *clientv3.Client) (*tor.Controller, error
...
+	if cfg.StaticSecret != nil && cfg.StaticSecret.Enabled {
+		seed = &cfg.StaticSecret.Seed
```
guggero, line 473: We should require the seed to have a minimum length to make sure it provides enough entropy. And we should stretch it to give even more entropy. We use `golang.org/x/crypto/scrypt` for that.
I think something like
```
scryptN = 8192
scryptR = 8
scryptP = 1
```
should provide a good tradeoff between startup speed and bruteforce-ability.

mint/mint.go
```diff
@@ -231,6 +249,55 @@ type VerificationParams struct {
 	TargetService string
 }
 
+// Generates a pseudo-random secret.
+func (m *Mint) generatePseudoRandomSecret(id *lsat.Identifier) ([lsat.SecretSize]byte, error) {
```
guggero, line 253: nit: call this `getDeterministicSecret`?

mint/mint.go
```diff
@@ -231,6 +249,55 @@ type VerificationParams struct {
...
+	}
+
+	copy(ret[0:], result[0:lsat.SecretSize])
+
+	return ret, nil
+}
+
+// Either returns the secret from SecretStore or creates a pseudo-random one.
+func (m *Mint) determineSecret(ctx context.Context, params *VerificationParams, id *lsat.Identifier) ([lsat.SecretSize]byte, error) {
+	if m.cfg.SeedForPseudoRandomness == nil {
```
guggero, line 274: nit: flip the condition around. Meaning `if ... != nil { return m.generatePseudoRandomSecret() }` as the branch is the path less often taken (or non-default path).

mint/mint.go
```diff
@@ -231,6 +249,55 @@ type VerificationParams struct {
...
+		return m.cfg.Secrets.GetSecret(
+			ctx, sha256.Sum256(params.Macaroon.Id()),
+		)
+	}
+	return m.generatePseudoRandomSecret(id)
+}
+
+// Creates a new secret (persistently stored or pseudo-random).
+func (m *Mint) createSecret(ctx context.Context, secretID [32]byte, id *lsat.Identifier) ([lsat.SecretSize]byte, SecretRemover, error) {
+	if m.cfg.SeedForPseudoRandomness == nil {
```
guggero, line 284: nit: same here, swap non-default and default paths for increased readability.

mint/mint.go
```diff
@@ -231,6 +249,55 @@ type VerificationParams struct {
...
+
+		secret, err := m.cfg.Secrets.NewSecret(ctx, secretID)
+
+		remover := func(ctx context.Context) error {
+			return m.cfg.Secrets.RevokeSecret(ctx, secretID)
+		}
+		return secret, remover, err
+	}
+
+	secret, err := m.generatePseudoRandomSecret(id)
```
guggero, line 294: Some unit tests around secret generation would be nice.

mint/mint.go
```diff
@@ -231,6 +249,55 @@ type VerificationParams struct {
 	TargetService string
 }
 
+// Generates a pseudo-random secret.
+func (m *Mint) generatePseudoRandomSecret(id *lsat.Identifier) ([lsat.SecretSize]byte, error) {
+	var ret [lsat.SecretSize]byte
+
+	mac := hmac.New(sha256.New, []byte(*m.cfg.SeedForPseudoRandomness))
```
guggero, line 256: Use the prepared and stretched secret here.

guggero (commented):
Great initiative, thank you for the PR!
I did a first pass review.

I think it's a nice idea to have a lightweight mode where you don't need `etcd`. And deterministically deriving the base macaroon secret with another HMAC is quite clever indeed.
The only usability/security issue I see with this is that you lose the ability to revoke a single token (can only revoke all of them by changing the static secret). And you don't have any idea how many tokens you have given out.
This could quite easily be fixed by creating a file that keeps track of all the IDs given out and checking if there is an entry when validating a macaroon. Deleting a token from the file would then invalidate it.
But maybe for most of the simpler use cases individual revocation is not needed.

The commit structure requires some more work, but I think over all this PR looks pretty good. Well done!

fiksn:
@guggero, please take another look. I believe I addressed all your comments. Nice that you pointed out key stretching, I was just thinking about PBKDF2 but thought that will be a pain to do. However scrypt API is really unbelievably simple. Not really sure about the parameters (perhaps this should be in constants?) and salt.

> Perhaps we need a new top-level option, for example backend-mode that is either etcd or lightweight. That would replace the staticSecret.Enabled and tor.Filestorage flags.

It's always a trade-off. But I believe if you say in the documentation "if you want to use aperture w/o etcd you need A and B" that is fine (since users then still have the option to just use A or B on their own). I want to leave that flexibility. It would be different if you needed more knobs, then I agree that some simple setting would be beneficial.

fiksn:
So I've just squashed the commits together (there is also one additional change that was related to error in cleanup when etcdClient is nil).

guggero:
Concept ACK, thanks for the changes!

Would be great if you could split this into smaller and easier to digest commits with logical changes to make a second round of review easier.
There are also a number of style nits that should still be addressed. Please take a look at https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md, as we try to adhere to those rules in all our projects.

We are a very small team and reviewing code thoroughly takes a lot of time. Therefore we need to heavily prioritize on PRs from external contributors and try to make sure they pass the minimal formal rules before we invest time into an in-depth review. I hope you understand.

guggero:
Closing due to inactivity.

PR #51, opened 2021-04-28
Remove token if payment failed and try again

guggero:
Related to https://github.com/lightninglabs/pool/issues/221.

This PR removes a pending LSAT if the payment for it is in a terminal failed state and tries a new payment. This removes the need for the user to manually remove the `lsat.token.pending` file from disk to try again.

### DIFF
--- a/lsat/client_interceptor.go
+++ b/lsat/client_interceptor.go
@@ -3,6 +3,7 @@ package lsat
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"regexp"
 	"sync"
@@ -58,6 +59,12 @@ var (
 	authHeaderRegex = regexp.MustCompile(
 		"LSAT macaroon=\"(.*?)\", invoice=\"(.*?)\"",
 	)
+
+	// errPaymentFailedTerminally is signaled by the payment tracking method
+	// to indicate a payment failed for good and will never change to a
+	// success state.
+	errPaymentFailedTerminally = errors.New("payment is in terminal " +
+		"failure state")
 )
 
 // ClientInterceptor is a gRPC client interceptor that can handle LSAT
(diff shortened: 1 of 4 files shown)


proxy/proxy_test.go
```diff
@@ -81,7 +81,11 @@ func TestProxyHTTP(t *testing.T) {
 		Addr:    testProxyAddr,
 		Handler: http.HandlerFunc(p.ServeHTTP),
 	}
-	go func() { _ = server.ListenAndServe() }()
+	go func() {
+		if err := server.ListenAndServe(); err != http.ErrServerClosed {
+			t.Errorf("Error serving on %s: %v", testProxyAddr, err)
```
Roasbeef, line 86: Perhaps fail the test all together in this case?

lsat/client_interceptor.go
```diff
@@ -408,6 +438,13 @@ func (i *ClientInterceptor) trackPayment(ctx context.Context, token *Token) erro
 			// time to complete.
 			case lnrpc.Payment_IN_FLIGHT:
 
+			// The payment is in a terminal failed state, it will
```
Roasbeef, line 441: 👍

lsat/client_interceptor.go
```diff
@@ -249,6 +256,29 @@ func (i *ClientInterceptor) handlePayment(iCtx *interceptContext) error {
...
+			iCtx.token = nil
+			if err := i.store.RemovePendingToken(); err != nil {
+				return fmt.Errorf("error removing pending "+
+					"token, cannot retry payment: %v", err)
+			}
+
+			// Let's try again by paying for the new token.
+			log.Infof("Retrying payment of LSAT token invoice")
+			var err error
+			iCtx.token, err = i.payLsatToken(
```
Roasbeef, line 273: Will this result in us continually retrying the payment if say, the user has no channels or one connected to a very distant edge?

Roasbeef (commented):
Great fix! Only lingering question is if this'll cause continual retries for a given call? AFAICT the answer is no, and it'll just do that retry per _new_ call.

proxy/proxy_test.go
wpaulino: As of 1.16, go emits a warning when `t.Fatal` is called within a goroutine.

lsat/client_interceptor.go
wpaulino: It seems that per RPC invocation, if there isn't a pending token, a new payment will be initiated for a new token, and if that fails, we'll try again. When a pending token exists however, a new payment for a new token is only attempted once.

wpaulino (approved):

proxy/proxy_test.go
guggero: Yes, you can't fail a test from a goroutine (or aren't supposed to, now enforced through the linter as @wpaulino pointed out). And it seemed a bit overkill to add an err chan to get the error out in this case. At least you see the error when running the test manually and know what's up.

lsat/client_interceptor.go
guggero: Almost correct. If there is no token at all, we initiate a new payment. That can only result in two outcomes: it succeeds and we continue, it fails (including timing out) and we error out. So there is no re-try for the case where we initiate the payment because of the timeout case (the payment could still be in-flight).
What changed is what happens if the user tries again: If we now see the payment is in a terminal state, we remove it and go into the "new payment" flow as described above.

tl;dr: there won't be an automatic re-try loop. but the user can manually try the same call again without needing to delete the token until eventually a payment succeeds.

guggero:
> Only lingering question is if this'll cause continual retries for a given call?

Interpreting this as an "LGTM if there is no re-try loop" and merging to unblock PRs in Loop and Pool.

PR #52, opened 2021-05-07
Fix "server closed the stream without sending trailers" error

guggero:
This PR adds a test to reproduce the elusive `server closed the stream without sending trailers` error that we first discovered through user feedback in https://github.com/lightninglabs/pool/issues/225.

Apparently this is a bug in the `net/http/httputil` package itself, so we can not really fix it but add a workaround.
The tl;dr is that sometimes, the HTTP/2 trailer fields aren't sent back to the client correctly when an error occurs on the proxied connection. We add our own HTTP transport implementation that fixes the problem.


guggero:
This could be somewhat related to https://github.com/golang/go/issues/34413. Maybe the backend closes the stream after sending an error. I'm not sure if I should invest more time into a repro scenario. Would be cool to fix something in golang though :)

Roasbeef:
Lint fails: 
```
proxy/proxy.go:363:22: resp.Header.Values undefined (type http.Header has no field or method Values) (typecheck)
		if len(resp.Header.Values(hdrGrpcStatus)) > 0 {
		                   ^
		                 ```

File: proxy/proxy.go [hunk 1]
```diff
@@ -329,11 +331,42 @@ func sendDirectResponse(w http.ResponseWriter, r *http.Request,
 	// so we can use that.
 	switch {
 	case strings.HasPrefix(r.Header.Get(hdrContentType), hdrTypeGrpc):
-		w.Header().Set("Grpc-Status", strconv.Itoa(int(codes.Internal)))
-		w.Header().Set("Grpc-Message", errInfo)
+		w.Header().Set(hdrGrpcStatus, strconv.Itoa(int(codes.Internal)))
```
Roasbeef, line 334: Any link handy to allow me to better understand whcih headers we're setting here and why?

proxy/proxy.go
```diff
@@ -329,11 +331,42 @@ func sendDirectResponse(w http.ResponseWriter, r *http.Request,
...
+// trailers aren't properly forwarded in some cases. We fix this by always
+// copying the Grpc-Status and Grpc-Message fields to the trailers, as those are
+// usually expected to be in the trailer fields.
+// Inspired by https://github.com/elazarl/goproxy/issues/408.
+func (l *trailerFixingTransport) RoundTrip(req *http.Request) (*http.Response,
+	error) {
+
+	resp, err := l.next.RoundTrip(req)
+	if resp != nil && len(resp.Trailer) == 0 {
+		if len(resp.Header.Values(hdrGrpcStatus)) > 0 {
```
Roasbeef, line 363: Could use a comment stating that the `OK` code is zero, so anything beyond that is actually considered an error: https://pkg.go.dev/google.golang.org/grpc/codes

proxy/proxy_test.go
```diff
@@ -195,6 +213,17 @@ func TestProxyGRPC(t *testing.T) {
 			runGRPCTest(t, tc)
 		})
 	}
+
+	for i := 0; i < 20; i++ {
```
Roasbeef, line 217: What's the goal w/ this added iteration of the test case? Given that the iteration index (or no other new information) is actually propagated into the test context.

proxy/proxy_test.go
```diff
@@ -327,38 +376,76 @@ func startBackendGRPC(grpcServer *grpc.Server) error {
...
+	for _, p := range tlsConf.NextProtos {
+		if p == "h2" {
+			haveNPN = true
+			break
+		}
+	}
+	if !haveNPN {
+		tlsConf.NextProtos = append(tlsConf.NextProtos, "h2")
+	}
+	tlsConf.NextProtos = append(tlsConf.NextProtos, "h2-14")
```
Roasbeef, line 423: Would this serve to resolve those issues we observed when we attempted to expose Aperture _directly_ over the load balancer?

proxy/testdata/hello.proto
```diff
@@ -3,12 +3,13 @@ syntax = "proto3";
 package proxy_test;
 
 service Greeter {
-    rpc SayHello (HelloRequest) returns (HelloReply) {}
-    rpc SayHelloNoAuth (HelloRequest) returns (HelloReply) {}
+    rpc SayHello (HelloRequest) returns (HelloReply);
+    rpc SayHelloNoAuth (HelloRequest) returns (HelloReply);
```
Roasbeef, line 7: Woah, new proto syntax change?

proxy/proxy.go
guggero: It's not a very good doc, but that's the best I could find: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

The `Grpc-Status` is the field that holds the numerical value for the gRPC specific states (e.g. `OK`, `INTERNAL`, `UNKNOWN`, `UNIMPLEMENTED`, `UNAVAILABLE` and so on). The `Grpc-Message` field contains the actual error message in case the gRPC status isn't `OK` (0). Those two fields should be sent as a HTTP/2 trailer field and not header field. That's what we tell the golang `http2` library by setting the `Trailer` field with the name of those two fields.

The `Content-Length` and `:status` are probably not necessary as they would be set anyway by the proxy. Maybe I should remove them as they are from trying out stuff.

guggero:
Argh, now there's a data race in the built in library...

Roasbeef (approved):
LGTM 🌛

PR #53, opened 2021-05-17
multi: Add gRPC price server option

ellemouton:
Adds ability to set a gRPC endpoint to hit for dynamic resource prices of a service. The backend service must then implement the gRPC service that will supply the prices for the various resources.

So this lets you, for example, have a service that matches on `stuff.com/blogs`, and then instead of charging the same price for each path (`blogs/1`, `blogs/2` etc) the backend server is queried with the path and the price for the specific resource is returned. This means that prices can be changed and resources can be added with different prices all without aperture needing to be aware of each individual resource. 

Relevant to #7.

I wrote an example server to test this branch out [here](https://github.com/ellemouton/aperturetest).


pricer/grpcPricer.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// NewGRPCPricer initialises a Pricer backed by a gRPC backend server.
+func NewGRPCPricer(cfg Config) (*GRPCPricer, error) {
+	var (
+		c   GRPCPricer
+		err error
+	)
+
+	var opt grpc.DialOption
```
guggero, line 39: nit: move into above `var` block.

pricer/pricer.go
```diff
@@ -0,0 +1,12 @@
+package pricer
+
+import "context"
+
+// Pricer is an interface used to query price data from a price provider.
+type Pricer interface {
+	// GetPrice should return the price for the given resource path.
```
guggero, line 7: Should mention it's the price in satoshis.

aperture.go
```diff
@@ -464,7 +464,10 @@ func createProxy(cfg *config, challenger *LndChallenger,
 }
 
 // cleanup closes the given server and shuts down the log rotator.
-func cleanup(etcdClient io.Closer, server io.Closer) {
+func cleanup(etcdClient io.Closer, server io.Closer, proxy io.Closer) {
```
guggero, line 467: nit: these two changes should be their own commit.

pricesrpc/rpc.proto
```diff
@@ -0,0 +1,17 @@
+syntax="proto3";
+
+package pricesrpc;
+
+option go_package = "github.com/lightninglabs/aperture/pricesrpc";
+
+service Prices {
+        rpc GetPrice(GetPriceRequest) returns (GetPriceResponse) {}
```
guggero, line 8: nit: can replace empty body with just a `;`.

pricer/defaultPricer.go
```diff
@@ -0,0 +1,29 @@
+package pricer
```
guggero, line 1: nit: Add the whole `pricer` package in its own commit?

sample-conf.yaml
```diff
@@ -83,7 +83,25 @@ services:
...
-    price: 1     
+    price: 1
+
+    # Whether or not a gRPC server is available to query price data from. If
+    # this option is set then the 'price' option is ignored.
+    usepricerserver: true
+
+    # Options to use for connection to the price serving gRPC server. If the
+    # 'usepricerserver' option is set then these config options must be set.
+    pricerconfig:
```
guggero, line 94: What do you think about the following hierarchy of options instead?
```yaml

price: 0
dynamicprice:
    enabled: true
    grpcaddress: "127.0.0.1:10010"
    insecure: false
    tlscertpath: "path-to-pricer-server-tls-cert/tls.cert"
```

That way we need an additional `nil` check for the `DynamicPrice` field but only introduce one new top-level field.
Just a thought, might be a bit easier to understand that way?

guggero (commented):
Great PR, thank you for the new feature!
Only nits and suggestions from my side :+1:

ellemouton:
Thanks for the review @guggero 🚀  I have updated it as suggested

proxy/proxy.go
```diff
@@ -107,16 +107,35 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	authLevel := target.AuthRequired(r)
 	switch {
 	case authLevel.IsOn():
-		if !p.authenticator.Accept(&r.Header, target.Name) {
+		if !p.authenticator.Accept(
```
guggero, line 110: nit: separate condition and method invocation?
For example:
```
acceptAuth := p.authenticator.Accept(
    ...
)
if !acceptAuth {
...
}
```

proxy/proxy.go
```diff
@@ -186,6 +224,12 @@ func (p *Proxy) UpdateServices(services []*Service) error {
 
 // Close cleans up the Proxy by closing any remaining open connections.
 func (p *Proxy) Close() error {
+	for _, s := range p.services {
```
guggero, line 227: We probably don't want to _not_ close the other pricers if one of them fails.
What we usually do in such a case is to declare a `var returnErr error` and set that if one of the errors is non-nil and then return that variable in the end. That way we only return the last error but at least close all of the items in the loop. Can add a log statement to make sure consecutive errors are at least seen.

proxy/proxy.go
```diff
@@ -128,7 +147,26 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 			if !ok {
-				p.handlePaymentRequired(w, r, target.Name, target.Price)
+				price, err := target.pricer.GetPrice(
+					r.Context(), r.URL.Path,
+				)
+				if err != nil {
+					prefixLog.Errorf("error getting"+
+						" resource price: %v", err)
```
guggero, line 155: We need to send a response here with `sendDirectResponse()`, probably an internal server error would be appropriate here.  Same in the `case` above.

guggero (commented):
Nice work on this! Latest iteration looks pretty good. Just a few more small things.
Going to add a second reviewer to this PR as well.

proxy/proxy.go
ellemouton: ah yes! good catch

ellemouton:
Cool, updated as suggested @guggero :) thanks for the review!

guggero (approved):
One last performance related issue, otherwise LGTM :tada:

proxy/proxy.go
```diff
@@ -105,18 +105,44 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Determine auth level required to access service and dispatch request
 	// accordingly.
 	authLevel := target.AuthRequired(r)
+
+	// Determine if the header contains the necessary authentication
+	// required for the given resource.
+	resourceName := target.ResourceName(r.URL.Path)
+	acceptAuth := p.authenticator.Accept(&r.Header, resourceName)
```
guggero, line 112: Oops, forgot to submit the comment...

I think this would now be called for every call, even static resources. This is probably too much of a performance hit as this might make one or more calls to `lnd`. We probably need to move the `Accept()` call inside the `switch` case again. Can add a comment to explain why it's in both `case` bodies.

ellemouton: oh yes of course! thanks for catching that!

Roasbeef:
Totally forgot about this, needs a rebase!

PR #54, opened 2021-05-31
proxy: remove manual trailer fields

guggero:
The last bugfix (#52) added some explicit header fields in an attempt of
fixing an issue with error responses. Unfortunately they weren't
strictly needed for the fix but ended up causing issues in a non-error
case.
This commit removes those header fields again and makes sure the "auth
header not found in response" header doesn't occur anymore.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -333,10 +333,6 @@ func sendDirectResponse(w http.ResponseWriter, r *http.Request,
 	case strings.HasPrefix(r.Header.Get(hdrContentType), hdrTypeGrpc):
 		w.Header().Set(hdrGrpcStatus, strconv.Itoa(int(codes.Internal)))
 		w.Header().Set(hdrGrpcMessage, errInfo)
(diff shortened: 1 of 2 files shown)


guggero:
In HTTP/2 you can send fields in the trailer section of a frame. In the `Trailer` header field you list all field names of the fields that are sent in the trailer instead of the header section. I think because we interfered with the default gRPC behavior (even though it was just _adding_ stuff to the `Trailer` header field), it caused the `WWW-Authenticate` header not to be sent correctly anymore. Which is where we place the LSAT challenge.

It's all my fault really, I shouldn't have included any changes in #52 that weren't strictly necessary. I tried out so many things to reproduce/fix that issue that in the end I wasn't sure anymore what actually fixed it so I left those header changes in. Thanks to the repro test we know now that both bugs are fixed with the current code version.

wpaulino (approved):

Roasbeef (approved):
LGTM 🌳

PR #56, opened 2021-06-29
Refactor to be run as library in integration tests

guggero:
Depends on #54.

Refactor aperture to be easily configurable and runnable as as a library in other projects. This will allow us to add aperture to our integration tests.


aperture.go
```diff
@@ -72,32 +72,32 @@ func Main() {
 	// TODO: Prevent from running twice.
 	err := run()
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
+		_, _ = fmt.Fprintln(os.Stderr, err)
```
ellemouton, line 75: does this addition do anything?

ellemouton (approved):
LGTM 🔥

aperture.go
guggero: Just removes a linter warning in my IDE. I try to clean them up whenever I see them, mostly to reduce my code OCD symptoms :joy:

go.mod
```diff
@@ -3,18 +3,18 @@ module github.com/lightninglabs/aperture
 go 1.15
 
 require (
-	github.com/btcsuite/btcd v0.21.0-beta.0.20201208033208-6bd4c64a54fa
+	github.com/btcsuite/btcd v0.21.0-beta.0.20210513141527-ee5896bad5be
```
wpaulino, line 6: We should be able to update our etcd client dependency as well and remove the replace workaround.

wpaulino (approved):

go.mod
guggero: Good point, I updated `lnd` to the version that bumps the `etcd` client version. Nice to finally get that replace removed.

PR #57, opened 2021-07-22
aperture: add servername to self-signed cert

thinkmassive:
Use the existing `servername` config parameter (also used with LetsEncrypt when `autocert = true`) as an extra domain to be added to self-signed certificates. This is useful for non-public endpoints, where autocert is not an option.


guggero (approved):
Nice, LGTM :100:

ellemouton (approved):
LGTM

PR #58, opened 2021-09-27
aperture: allow command line flags and add config file

carlaKC:
This PR updates aperture to allow specifying configuration options on the command line and adds an optional `configfile` flag to point to a custom location (rather than the default `aperture.yaml`).

This change relaxes our requirement that the default `aperture.yaml` exists, since we could have all of our options specified inline. It also tightens up the validation we require on our config to ensure that we start up with the minimum required config.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -102,8 +102,7 @@ func run() error {
 	}
 
 	// Next, parse configuration file and set up logging.
-	configFile := filepath.Join(apertureDataDir, defaultConfigFilename)
-	cfg, err := getConfig(configFile)
+	cfg, err := getConfig()
 	if err != nil {
 		return fmt.Errorf("unable to parse config file: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


carlaKC:
@ellemouton @guggero will you take a look? Don't have perms to request review.

aperture.go
```diff
@@ -316,6 +317,12 @@ func getConfig(configFile string) (*Config, error) {
 		return nil, err
 	}
 
+	// Finally, parse the remaining command line options again to ensure
+	// they take precedence.
+	if _, err := flags.Parse(cfg); err != nil {
+		return nil, err
```
guggero, line 352: I think we need to check the error returned here or where this function is called. If I run `aperture --help` I get the help text twice.

guggero (approved):
Nice, LGTM :100:

aperture.go
ellemouton: yeah. In lnd we do this:

```
loadedConfig, err := lnd.LoadConfig(shutdownInterceptor)
if e, ok := err.(*flags.Error); !ok || e.Type != flags.ErrHelp {
			// Print error if not due to help request.
			_, _ = fmt.Fprintln(os.Stderr, err)
			os.Exit(1)
		}

		// Help was requested, exit normally.
		os.Exit(0)
}
```

aperture.go
```diff
@@ -307,15 +306,44 @@ func fileExists(name string) bool {
...
-	b, err := ioutil.ReadFile(configFile)
-	if err != nil {
+	if _, err := flags.Parse(cfg); err != nil {
 		return nil, err
 	}
-	err = yaml.Unmarshal(b, cfg)
-	if err != nil {
+
+	// If a custom config file is provided, we require that it exists.
+	var mustExist bool
```
ellemouton, line 318: surely we always require that one or the other exists?

ellemouton (approved):
Nice! This is gonna be super useful. 🚀

aperture.go
carlaKC: Ah yeah, I always forget this! 

Gone with a slightly different solution here,wrapping the parser error (then unwrapping), but essentially the same.

carlaKC:
Diff updating `--help` handling [here](https://github.com/lightninglabs/aperture/compare/8227350505c050c14e24f6efeabf4f1d0f775cb9..0b8128232d54e91962a8c8d42fa019452d5fdfd6).

aperture.go
carlaKC: Now that we allow command line flags, you could _theoretically_ specify everything on the command line. In that case, we don't want to fail if somebody gives us all the flags we need on the cli but the default location file doesn't exit. Whereas when you specifically set `--configfile`, I think it's fair to expect it to be there?

PR #59, opened 2021-10-05
multi: add base dir to aperture

carlaKC:
This PR adds a base dir to aperture for move convenient file placement. 

There's always some thinking to so with our various file placement flags (specifically `--configfile` and `--basedir`), gone with the following logic: 
* Default to `AppDir` if nothing is set
* If only `--configfile` is set, use `AppDir` for files, use custom config location and require that it exists
* If only `--basedir` is set, look for `aperture.yaml` in `basedir` but do not require it
* If `--configfile` and `--basedir` are set, use `basedir` for files and require `--configfile` exists

Reason for allowing a `--configfile` and `--basedir` is that custom config file allows you to set the name of your config file as well as location.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -212,7 +212,7 @@ func (a *Aperture) Start(errChan chan error) error {
 		a.httpsServer.Handler = h2c.NewHandler(handler, &http2.Server{})
 	} else {
 		a.httpsServer.TLSConfig, err = getTLSConfig(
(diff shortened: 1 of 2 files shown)


carlaKC:
@guggero @ellemouton PTAL 🙏

guggero (approved):
Niiice, LGTM :tada:

ellemouton (approved):

PR #65, opened 2022-01-28
grpc update draft

bhandras:


guggero:
Not needed since workaround in https://github.com/lightninglabs/aperture/pull/66 seems to be working.

PR #67, opened 2022-02-02
aperture: connect REST proxy correctly when insecure is set

guggero:
Fixes an issue where the REST proxy couldn't connect to itself when running with `insecure: true` in the main config file.


thinkmassive (approved):
Most excellent!

ellemouton (approved):

PR #68, opened 2022-02-08
aperture: hashmail keepalive enforcement policy

ellemouton:
Lower the MinTime frequency that the client is allowed to send pings to
1 min instead of the default of 5 mins.


ellemouton:
cc @guggero ^ (i cant request review in this repo it seems)

guggero (approved):

PR #70, opened 2022-02-11
aperture: add prometheus config nil check

ellemouton:
check if prometheus config is nil before using it

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -181,9 +181,10 @@ func (a *Aperture) Start(errChan chan error) error {
 	var err error
 
 	// Start the prometheus exporter.
-	if err := StartPrometheusExporter(a.cfg.Prometheus); err != nil {
-		return fmt.Errorf("unable to start the prometheus exporter: "+
-			"%v", err)
+	err = StartPrometheusExporter(a.cfg.Prometheus)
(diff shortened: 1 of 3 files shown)


ellemouton:
cc @guggero 

(sorry about this. I assumed the build fail on the og pr was due to the usual data race reasons.)

aperture.go
```diff
@@ -180,10 +180,13 @@ func NewAperture(cfg *Config) *Aperture {
 func (a *Aperture) Start(errChan chan error) error {
 	var err error
 
-	// Start the prometheus exporter.
-	if err := StartPrometheusExporter(a.cfg.Prometheus); err != nil {
-		return fmt.Errorf("unable to start the prometheus exporter: "+
-			"%v", err)
+	if a.cfg.Prometheus != nil {
```
guggero, line 183: Should we instead add a `DefaultConfig()` function and use that in `aperture.go#L366`?

ellemouton: Can do. Or can add a  "NewConfig" func for now since we dont really have other defaults and just want to initialise the pointer variables. wdyt?

guggero (approved):

PR #74, opened 2022-03-31
multi: add mailbox read counter

ellemouton:
In this commit, we add a mailbox-read-count metric. This will be
incremented each time a mailbox with an _odd_ stream ID is read from.
We do this because we assume that a full duplex connection is being used
meaning that there will be 2 streams that have a matching ID except for
the last byte. And so to avoid duplicating the data, we only record the
odd streams. We also assume that for every read, there will be a write
and so we only record the reads.

### DIFF
--- a/prometheus.go
+++ b/prometheus.go
@@ -8,12 +8,28 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+const streamIDLabel = "streamID"
+
 var (
 	// mailboxCount tracks the current number of active mailboxes.
 	mailboxCount = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: "hashmail",
 		Name:      "mailbox_count",
 	})
+
+	// mailboxReadCount counts each time a mailbox pair is being used.
+	// A session consists of a bidirectional stream each using a mailbox
+	// with an ID that overlaps for the first 63 bytes and differ for the
+	// last bit. So in order to obtain accurate data about a specific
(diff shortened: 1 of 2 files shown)


ellemouton:
cc @guggero 

im not 100% sure this approach is what we want since it makes the assumption that mailboxes are used in pairs where the pairs have 63 bytes of matching IDs and 1 byte not. But this is LNC specific so not sure we should make this assumption here. Was trying to avoid the need to divide by 2 later. but maybe that is actually better?

ellemouton:
Thanks for the review @guggero!
updated to only use the first 16 bytes as the base ID 👍

prometheus.go
```diff
@@ -8,12 +8,28 @@ import (
...
 	// mailboxCount tracks the current number of active mailboxes.
 	mailboxCount = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace: "hashmail",
 		Name:      "mailbox_count",
 	})
+
+	// mailboxReadCount counts each time a mailbox pair is being used.
+	// A session consists of a bidirectional stream each using a mailbox
+	// with an ID that overlaps for the first 63 bytes and differ for the
+	// last byte. So in order to obtain accurate data about a specific
```
guggero, line 23: micro nit: It's only the last "bit", not last byte.

guggero (approved):
utACK, LGTM :taco:

prometheus.go
ellemouton: indeed! fixed 👍

Issue #55, opened 2021-06-16
BOS scores endpoint in the testnet environment returning 404


lingmann:
The BOS scores endpoint seems to be down and has been returning a 404 since at least yesterday. FYI!

```text
$ curl -k -v https://test.swap.lightning.today:11010/availability/v1/btc.json
*   Trying 35.160.63.96...
* TCP_NODELAY set
* Connected to test.swap.lightning.today (35.160.63.96) port 11010 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: CN=test.swap.lightning.today
*  start date: Jul 26 00:00:00 2020 GMT
*  expire date: Aug 26 12:00:00 2021 GMT
*  issuer: C=US; O=Amazon; OU=Server CA 1B; CN=Amazon
*  SSL certificate verify ok.
> GET /availability/v1/btc.json HTTP/1.1
> Host: test.swap.lightning.today:11010
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Wed, 16 Jun 2021 20:31:23 GMT
< Content-Length: 19
<
404 page not found
* Connection #0 to host test.swap.lightning.today left intact
* Closing connection 0
```

guggero:
The example page currently isn't hosted by us anymore and we should remove that section from the README.

[...]

PR #75, opened 2022-04-20
#55 Removing demo section in README.md

hieblmi:
As suggested here https://github.com/lightninglabs/aperture/issues/55 the demo section is no longer functional hence removing it.


guggero (approved):
Way overdue, thanks for the update.

PR #76, opened 2022-04-26
multi: profile server

ellemouton:
This commit adds a config option that can be set in order to spin up a
pprof profile server on the given port.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -37,6 +37,9 @@ import (
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/protobuf/encoding/protojson"
 	"gopkg.in/yaml.v2"
+
+	// Blank import to set up profiling HTTP handlers.
+	_ "net/http/pprof"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


ellemouton:
@guggero & @thinkmassive , I am unable to request reviewers on this repo it seems. So just pinging you guys instead :) 
Im currently trying to find a possible memory leak but would be useful to be able to profile our prod mailbox servers

aperture.go
```diff
@@ -185,6 +188,24 @@ func (a *Aperture) Start(errChan chan error) error {
...
+			return fmt.Errorf("the profile port must be between " +
+				"1024 and 65535")
+		}
+
+		go func() {
+			http.Handle("/", http.RedirectHandler(
+				"/debug/pprof", http.StatusSeeOther,
+			))
+
+			listenAddr := fmt.Sprintf(":%d", a.cfg.ProfilePort)
```
guggero, line 203: This needs to be `localhost:%d`, otherwise it will listen on all IP addresses!

guggero (changes requested):

Issue #80, opened 2022-09-21
Ability to deal in relative times with aperture created lsats


bucko13:
I noticed some breadcrumbs in the existing code that indicate there was the intention to build in some support for this but best I can tell they are mainly just placeholders for now. 

token.go just sets a `TimeCreated` value when [parsing from the challenge](https://github.com/lightninglabs/aperture/blob/master/lsat/token.go#L59-L63) but that's just set to `time.Now()`. As best I can tell, this isn't really persisted anywhere "permanent" on the lsat itself. Then there is [this code](https://github.com/lightninglabs/aperture/blob/master/lsat/token.go#L59-L63) that sets an `IsValid` method on the `Token`. It's only a placeholder now that always returns true. It's not clear to me how the `TimeCreated` value on the token serves any use if it's not encoded directly in the macaroon. I might be missing where this is persisted though. 

It seems to me like this would be pretty valuable functionality to have. Without the ability to use aperture as a library, the only way to add caveats via aperture as the author is hardcoded in the config file. As far as I can tell, this makes relative caveat restrictions like an expiration date impossible. The [sample config shows](https://github.com/lightninglabs/aperture/blob/master/sample-conf.yaml#L112) how to add an absolute expiration with a `valid_until` constraint, but I can't see how given the current tooling you could do something like "valid_for=xx_days", which feels much more valuable and flexible to me. 

Is there some way to achieve this that I'm missing? If not, what are the chances of getting this support in aperture?

guggero:
I don't see how a relative `valid_for` would work. When would that start counting? At the moment you add that caveat? Then you could just as well create an absolute `valid_until` field, which is basically `now+x days`. `lnd` does understand and validate the `valid_until` field, we just never added support for it in Aperture. But would be quite trivial to add.

bucko13:
The problem with `valid_until` is the way that aperture configs work (again, as far as I'm aware). Since they're static, you'd have to constantly keep updating it to useful. If i set that config to 1 year from now, each new lsat that's issued becomes more and more restrictive. 

What I think makes sense is to do something similar to what it looks like lnd does with `macaroontimeout` where in the config you set the timeout, and then the minter/bakery will add the caveat based on that number of seconds. I actually hacked together a version of this and it seems to work. I'll put a PR up in a bit for reference.

PR #81, opened 2022-09-22
Timeout Caveat Support

bucko13:
This is a proof of concept to address #80. It is loosely inspired by `macaroontimeout` and [`TimeoutConstraint`](https://github.com/lightningnetwork/lnd/blob/28ea2736a04566b06c38533e5e48e1b8d185a49b/macaroons/constraints.go#L74-L81) in lnd. 

The way the api for this is supposed to work is that if you want your minted lsats to have an "expiration" for specific services, you set a timeout property in your aperture config at the service level (same as `capabilities` and `constraints`). This should be a value that equates to the number of seconds after minting that the lsat should expire. When the lsat is minted, this value will be consumed and the unix timestamp for that number of seconds from that moment will be added as the value for a caveat with the condition `[service_name]_timeout`.

I've tested this out in an application I'm working on for which this is a feature requirement. If this isn't the preferred approach, I'd be happy to collaborate to find one that works or do what needs to be done to make this an official part of aperture.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -662,6 +662,7 @@ func createProxy(cfg *Config, challenger *LndChallenger,
 		Challenger:     challenger,
 		Secrets:        newSecretStore(etcdClient),
 		ServiceLimiter: newStaticServiceLimiter(cfg.Services),
+		Now:            time.Now,
 	})
 	authenticator := auth.NewLsatAuthenticator(minter, challenger)
 
--- a/lsat/satisfier.go
+++ b/lsat/satisfier.go
@@ -2,7 +2,9 @@ package lsat
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
+	"time"
 )
 
 // Satisfier provides a generic interface to satisfy a caveat based on its
@@ -79,7 +81,9 @@ func NewServicesSatisfier(targetService string) Satisfier {
 
 // NewCapabilitiesSatisfier implements a satisfier to determine whether the
 // target capability for a service is authorized for a given LSAT.
-func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier {
+func NewCapabilitiesSatisfier(service string,
+	targetCapability string) Satisfier {
+
 	return Satisfier{
 		Condition: service + CondCapabilitiesSuffix,
 		SatisfyPrevious: func(prev, cur Caveat) error {
@@ -112,6 +116,65 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
 			}
 			return fmt.Errorf("target capability %v not authorized",
 				targetCapability)
+		},
+	}
+}
+
+// NewTimeoutSatisfier checks if an LSAT is expired or not. The Satisfier takes
+// a service name to set as the condition prefix and currentTimestamp to
+// compare against the expiration(s) in the caveats. The expiration time is
+// retrieved from the caveat values themselves. The satisfier will also make
+// sure that each subsequent caveat of the same condition only has increasingly
+// strict expirations.
+func NewTimeoutSatisfier(service string, now func() time.Time) Satisfier {
+	return Satisfier{
+		Condition: service + CondTimeoutSuffix,
+		SatisfyPrevious: func(prev, cur Caveat) error {
+			prevValue, err := strconv.ParseInt(prev.Value, 10, 64)
+			if err != nil {
+				return fmt.Errorf("error parsing previous "+
+					"caveat value: %w", err)
+			}
+
+			currValue, err := strconv.ParseInt(cur.Value, 10, 64)
+			if err != nil {
+				return fmt.Errorf("error parsing caveat "+
+					"value: %w", err)
+			}
+
+			prevTime := time.Unix(prevValue, 0)
+			currTime := time.Unix(currValue, 0)
+
+			// Satisfier should fail if a previous timestamp in the
+			// list is earlier than ones after it b/c that means
+			// they are getting more permissive.
+			if prevTime.Before(currTime) {
+				return fmt.Errorf("%s caveat violates "+
+					"increasing restrictiveness",
+					service+CondTimeoutSuffix)
+			}
+
+			return nil
+		},
+		SatisfyFinal: func(c Caveat) error {
+			expirationTimestamp, err := strconv.ParseInt(
+				c.Value, 10, 64,
+			)
+			if err != nil {
+				return fmt.Errorf("caveat value not a valid "+
+					"integer: %v", err)
+			}
+
+			expirationTime := time.Unix(expirationTimestamp, 0)
+
+			// Make sure that the final relevant caveat is not
+			// passed the current date/time.
+			if now().Before(expirationTime) {
+				return nil
+			}
+
+			return fmt.Errorf("not authorized to access " +
+				"service. LSAT has expired")
 		},
 	}
 }
--- a/lsat/satisfier_test.go
+++ b/lsat/satisfier_test.go
@@ -1,0 +1,87 @@
+package lsat
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestTimeoutSatisfier tests that the Timeout Satisfier implementation behaves
+// as expected and correctly accepts or rejects calls based on if the
+// timeout has been reached or not.
+func TestTimeoutSatisfier(t *testing.T) {
+	t.Parallel()
+
+	now := int64(0)
+
+	var tests = []struct {
+		name           string
+		timeouts       []int64
+		expectFinalErr bool
+		expectPrevErr  bool
+	}{
+		{
+			name:     "current time is before expiration",
+			timeouts: []int64{now + 1000},
+		},
+		{
+			name: "time passed is greater than " +
+				"expiration",
+			timeouts:       []int64{now - 1000},
+			expectFinalErr: true,
+		},
+		{
+			name: "successive caveats are increasingly " +
+				"restrictive and not yet expired",
+			timeouts: []int64{now + 1000, now + 500},
+		},
+		{
+			name: "latter caveat is less restrictive " +
+				"then previous",
+			timeouts:      []int64{now + 500, now + 1000},
+			expectPrevErr: true,
+		},
+	}
+
+	var (
+		service   = "restricted"
+		condition = service + CondTimeoutSuffix
+		satisfier = NewTimeoutSatisfier(service, func() time.Time {
+			return time.Unix(now, 0)
+		})
+	)
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			var prev *Caveat
+			for _, timeout := range test.timeouts {
+				caveat := NewCaveat(
+					condition, fmt.Sprintf("%d", timeout),
+				)
+
+				if prev != nil {
+					err := satisfier.SatisfyPrevious(
+						*prev, caveat,
+					)
+					if test.expectPrevErr {
+						require.Error(t, err)
+					} else {
+						require.NoError(t, err)
+					}
+				}
+
+				err := satisfier.SatisfyFinal(caveat)
+				if test.expectFinalErr {
+					require.Error(t, err)
+				} else {
+					require.NoError(t, err)
+				}
+
+				prev = &caveat
+			}
+		})
+	}
+}
--- a/lsat/service.go
+++ b/lsat/service.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 )
 
 const (
@@ -15,6 +16,10 @@ const (
 	// capabilities caveat. For example, the condition of a capabilities
 	// caveat for a service named `loop` would be `loop_capabilities`.
 	CondCapabilitiesSuffix = "_capabilities"
+
+	// CondTimeoutSuffix is the condition suffix used for a service's
+	// timeout caveat.
+	CondTimeoutSuffix = "_valid_until"
 )
 
 var (
@@ -127,5 +132,21 @@ func NewCapabilitiesCaveat(serviceName string, capabilities string) Caveat {
 	return Caveat{
 		Condition: serviceName + CondCapabilitiesSuffix,
 		Value:     capabilities,
+	}
+}
+
+// NewTimeoutCaveat creates a new caveat that will result in a macaroon being
+// valid for numSeconds after the current time.
+func NewTimeoutCaveat(serviceName string, numSeconds int64,
+	now func() time.Time) Caveat {
+
+	var (
+		macaroonTimeout = time.Duration(numSeconds) * time.Second
+		requestTimeout  = now().Add(macaroonTimeout)
+	)
+
+	return Caveat{
+		Condition: serviceName + CondTimeoutSuffix,
+		Value:     strconv.FormatInt(requestTimeout.Unix(), 10),
 	}
 }
--- a/mint/mint.go
+++ b/mint/mint.go
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/lightninglabs/aperture/lsat"
 	"github.com/lightningnetwork/lnd/lntypes"
@@ -60,6 +61,10 @@ type ServiceLimiter interface {
 	// enforces additional constraints on a particular service/service
 	// capability.
 	ServiceConstraints(context.Context, ...lsat.Service) ([]lsat.Caveat, error)
+
+	// ServiceTimeouts returns the timeout caveat for each service. This
+	// will determine if and when service access can expire.
+	ServiceTimeouts(context.Context, ...lsat.Service) ([]lsat.Caveat, error)
 }
 
 // Config packages all of the required dependencies to instantiate a new LSAT
@@ -76,6 +81,9 @@ type Config struct {
 	// ServiceLimiter provides us with how we should limit a new LSAT based
 	// on its target services.
 	ServiceLimiter ServiceLimiter
+
+	// Now returns the current time.
+	Now func() time.Time
 }
 
 // Mint is an entity that is able to mint and verify LSATs for a set of
@@ -210,10 +218,15 @@ func (m *Mint) caveatsForServices(ctx context.Context,
 	if err != nil {
 		return nil, err
 	}
+	timeouts, err := m.cfg.ServiceLimiter.ServiceTimeouts(ctx, services...)
+	if err != nil {
+		return nil, err
+	}
 
 	caveats := []lsat.Caveat{servicesCaveat}
 	caveats = append(caveats, capabilities...)
 	caveats = append(caveats, constraints...)
+	caveats = append(caveats, timeouts...)
 	return caveats, nil
 }
 
@@ -269,6 +282,8 @@ func (m *Mint) VerifyLSAT(ctx context.Context, params *VerificationParams) error
 		caveats = append(caveats, caveat)
 	}
 	return lsat.VerifyCaveats(
-		caveats, lsat.NewServicesSatisfier(params.TargetService),
+		caveats,
+		lsat.NewServicesSatisfier(params.TargetService),
+		lsat.NewTimeoutSatisfier(params.TargetService, m.cfg.Now),
 	)
 }
--- a/mint/mint_test.go
+++ b/mint/mint_test.go
@@ -5,8 +5,10 @@ import (
 	"crypto/sha256"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/lightninglabs/aperture/lsat"
+	"github.com/stretchr/testify/require"
 	"gopkg.in/macaroon.v2"
 )
 
@@ -27,6 +29,7 @@ func TestBasicLSAT(t *testing.T) {
 		Secrets:        newMockSecretStore(),
 		Challenger:     newMockChallenger(),
 		ServiceLimiter: newMockServiceLimiter(),
+		Now:            time.Now,
 	})
 
 	// Mint a basic LSAT which is only able to access the given service.
@@ -46,7 +49,7 @@ func TestBasicLSAT(t *testing.T) {
 
 	// It should not be able to access an unknown service.
 	unknownParams := params
-	unknownParams.TargetService = "uknown"
+	unknownParams.TargetService = "unknown"
 	err = mint.VerifyLSAT(ctx, &unknownParams)
 	if !strings.Contains(err.Error(), "not authorized") {
 		t.Fatal("expected LSAT to not be authorized")
@@ -63,6 +66,7 @@ func TestAdminLSAT(t *testing.T) {
 		Secrets:        newMockSecretStore(),
 		Challenger:     newMockChallenger(),
 		ServiceLimiter: newMockServiceLimiter(),
+		Now:            time.Now,
 	})
 
 	// Mint an admin LSAT by not including any services.
@@ -92,6 +96,7 @@ func TestRevokedLSAT(t *testing.T) {
 		Secrets:        newMockSecretStore(),
 		Challenger:     newMockChallenger(),
 		ServiceLimiter: newMockServiceLimiter(),
+		Now:            time.Now,
 	})
 
 	// Mint an LSAT and verify it.
@@ -128,6 +133,7 @@ func TestTamperedLSAT(t *testing.T) {
 		Secrets:        newMockSecretStore(),
 		Challenger:     newMockChallenger(),
 		ServiceLimiter: newMockServiceLimiter(),
+		Now:            time.Now,
 	})
 
 	// Mint a new LSAT and verify it is valid.
@@ -175,6 +181,7 @@ func TestDemotedServicesLSAT(t *testing.T) {
 		Secrets:        newMockSecretStore(),
 		Challenger:     newMockChallenger(),
 		ServiceLimiter: newMockServiceLimiter(),
+		Now:            time.Now,
 	})
 
 	unauthorizedService := testService
@@ -224,4 +231,67 @@ func TestDemotedServicesLSAT(t *testing.T) {
 	if !strings.Contains(err.Error(), "not authorized") {
 		t.Fatal("expected macaroon to be invalid")
 	}
+}
+
+// TestExpiredServicesLSAT asserts the behavior of the Timeout caveat.
+func TestExpiredServicesLSAT(t *testing.T) {
+	t.Parallel()
+
+	initialTime := int64(1000)
+	mockTime := newMockTime(initialTime)
+
+	ctx := context.Background()
+	mint := New(&Config{
+		Secrets:        newMockSecretStore(),
+		Challenger:     newMockChallenger(),
+		ServiceLimiter: newMockServiceLimiter(),
+		Now:            mockTime.now,
+	})
+
+	// Mint a new lsat for accessing a test service.
+	mac, _, err := mint.MintLSAT(ctx, testService)
+	require.NoError(t, err)
+
+	authorizedParams := VerificationParams{
+		Macaroon:      mac,
+		Preimage:      testPreimage,
+		TargetService: testService.Name,
+	}
+
+	// It should be able to access the service if no timeout caveat added.
+	require.NoError(t, mint.VerifyLSAT(ctx, &authorizedParams))
+
+	// Add a timeout caveat that expires in the future.
+	timeout := lsat.NewTimeoutCaveat(testService.Name, 1000, mockTime.now)
+	require.NoError(t, lsat.AddFirstPartyCaveats(mac, timeout))
+
+	// Make sure that the LSAT is still valid after timeout is added since
+	// the timeout has not yet been reached.
+	require.NoError(t, mint.VerifyLSAT(ctx, &authorizedParams))
+
+	// Force time to pass such that the LSAT should no longer be valid.
+	mockTime.setTime(initialTime + 1001)
+
+	// Assert that the LSAT is no longer valid due to the timeout being
+	// reached.
+	err = mint.VerifyLSAT(ctx, &authorizedParams)
+	require.Contains(t, err.Error(), "not authorized")
+}
+
+type mockTime struct {
+	time time.Time
+}
+
+func newMockTime(initialTime int64) *mockTime {
+	return &mockTime{
+		time: time.Unix(initialTime, 0),
+	}
+}
+
+func (mt *mockTime) now() time.Time {
+	return mt.time
+}
+
+func (mt *mockTime) setTime(timestamp int64) {
+	mt.time = time.Unix(timestamp, 0)
 }
--- a/mint/mock_test.go
+++ b/mint/mock_test.go
@@ -73,6 +73,7 @@ func newMockSecretStore() *mockSecretStore {
 type mockServiceLimiter struct {
 	capabilities map[lsat.Service]lsat.Caveat
 	constraints  map[lsat.Service][]lsat.Caveat
+	timeouts     map[lsat.Service]lsat.Caveat
 }
 
 var _ ServiceLimiter = (*mockServiceLimiter)(nil)
@@ -81,6 +82,7 @@ func newMockServiceLimiter() *mockServiceLimiter {
 	return &mockServiceLimiter{
 		capabilities: make(map[lsat.Service]lsat.Caveat),
 		constraints:  make(map[lsat.Service][]lsat.Caveat),
+		timeouts:     make(map[lsat.Service]lsat.Caveat),
 	}
 }
 
@@ -108,6 +110,20 @@ func (l *mockServiceLimiter) ServiceConstraints(ctx context.Context,
 			continue
 		}
 		res = append(res, constraints...)
+	}
+	return res, nil
+}
+
+func (l *mockServiceLimiter) ServiceTimeouts(ctx context.Context,
+	services ...lsat.Service) ([]lsat.Caveat, error) {
+
+	res := make([]lsat.Caveat, 0, len(services))
+	for _, service := range services {
+		timeouts, ok := l.timeouts[service]
+		if !ok {
+			continue
+		}
+		res = append(res, timeouts)
 	}
 	return res, nil
 }
--- a/proxy/service.go
+++ b/proxy/service.go
@@ -72,6 +72,12 @@ type Service struct {
 	// the file is sent encoded as base64.
 	Headers map[string]string `long:"headers" description:"Header fields to always pass to the service"`
 
+	// Timeout is an optional value that indicates in how many seconds the
+	// service's caveat should time out relative to the time of creation. So
+	// if a value of 100 is set, then the timeout will be 100 seconds
+	// after creation of the LSAT.
+	Timeout int64 `long:"timeout" description:"An integer value that indicates the number of seconds until the service access expires"`
+
 	// Capabilities is the list of capabilities authorized for the service
 	// at the base tier.
 	Capabilities string `long:"capabilities" description:"A comma-separated list of the service capabilities authorized for the base tier"`
--- a/services.go
+++ b/services.go
@@ -2,6 +2,7 @@ package aperture
 
 import (
 	"context"
+	"time"
 
 	"github.com/lightninglabs/aperture/lsat"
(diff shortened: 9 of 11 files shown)


bucko13:
For reference I have a PR that adds a satisfier for validating this caveat on LSATs for the service I'm aiming to use this for. See `IsExpiredSatisfier` [here](https://github.com/stakwork/sphinx-meme/blob/36433dd4c76bf3d75d0d526d9223a24b00503579/lsat/satisfier.go#L87)

ellemouton (commented):
Hi @bucko13, thanks for the PR!

Some questions I have: why not just use the existing macaroon timeout caveat functions as is done in `lncli`? And also, I think that if we add specific conditions to aperture itself, then we also need to add the checks for those conditions in aperture. Because if I understand correctly, this PR just adds a new condition to the lsat but expects the downstream server to verify that a request is valid/invalid given the condition?

bucko13:
@ellemouton im not sure I totally understand. Is aperture used in lncli or between access from lncli and lnd? My understanding is that this is a proxy server used in a totally different type of environment (more like between client server) than between more of a “master” admin (lncli) talking with the tool you want to control access to (lnd). In fact, I based this caveat tooling on what I [assume is used in lncli](https://github.com/lightningnetwork/lnd/blob/28ea2736a04566b06c38533e5e48e1b8d185a49b/macaroons/constraints.go#L74-L81).

ellemouton:
@bucko13 , I just meant that as an example: like `lncli` uses a standard caveat name for the timeout that it gets from the macaroon lib. That way lnd can use the existing expiry check from the lib too.

With how this PR is currently - how I understand it is that it is adding a completely new caveat name to the macaroon and then when a request comes through with a macaroon that has that caveate, aperture itself is not checking the validity of the caveat but is rather depending on the backend service to know how to do that?

bucko13:
Ah, yeah sure I can change the name. Everything else still basically needs to be done after the caveat goes through as far as I understand the way aperture is constructed. Aperture right now can't verify that all of the services that you add on are valid, that has to be done after the base validation is done and the request has been passed through. 

[Here's an example](https://github.com/stakwork/sphinx-meme/blob/master/lsat/lsat.go#L129) in a tech stack where I'm trying to use aperture as a proxy, where I have a custom "sphinx-meme" service and all the limitations associated with that for requests. Aperture has now way to use the restrictions on its own, so that's just done in sphinx meme after the request has been proxied. 

The same would more or less have to happen with this timeout setting whether or not it's named differently, though it does make it easier to use the standard lib for leveraging it (as lnd presumably does). It's pretty simple to implement the satisfier though: [48 lines for the satisfier](https://github.com/stakwork/sphinx-meme/blob/master/lsat/lsat.go#L129) and then pass it in to a verifier like this: `VerifyCaveats(caveats, NewUploadSizeSatisfier(int64(handler.Size)), IsExpiredSatisfier(time.Now().Unix()))`.

Do you happen to know what I would need to change the name to so that it follows the standard? I've seen a few. There's [this library](https://www.npmjs.com/package/macaroons.js#verifying-macaroons-with-caveats) that was based on the reference (but has diverged since I think), but they use `time` in their examples. In the [go reference](https://github.com/go-macaroon/macaroon) I don't see anything obviously documented but I could be missing it.

bucko13:
@ellemouton I just wanted to bump this one more time to see if it's something that could be supported in the official aperture repo so we don't have to rely on a fork. 

I'd like to try and understand better what the main concern with this approach is. If it's simply that the external service needs to validate the caveat on its own, I can understand why that might be a problem however I'd just point out that as best I can see this is more or less how aperture is expected to function in most cases. 

Consider one of the examples in the lsat.tech [service constraints example](https://lsat.tech/macaroons#service-constraints) where there's a caveat that looks like ` loop_out_monthly_volume_sats = 200000000`. As best as I can tell, there's no way that aperture can know if this volume has been surpassed or not and it will have to be checked by the service on the backend anyway. In fact, for aperture to have any sort of flexibility most of the validation beyond just restricting access to specific paths is going to have to be done _after_ aperture has done a surface level validation of the macaroon. 

That said, I see no reason why I couldn't build on this PR to also add support to also validate the `timeout` validation (which incidentally as I mentioned [you already do in lncli](https://github.com/lightningnetwork/lnd/blob/28ea2736a04566b06c38533e5e48e1b8d185a49b/macaroons/constraints.go#L74-L81)). The only reason I didn't is because I thought it would be easier to keep the scope as reduced as possible and since this is how other caveats seem to work already it seemed reasonable to me. If I did add the validation as well, I'd like to try and get as clear as possible of an idea of what the expectations are for something like this. If it makes sense to you, I could simply move the [satisfier](https://github.com/stakwork/sphinx-meme/pull/10/files#diff-28ef562dd4a2fb366dc11e80f6daa38220b2d132d2878c752d2f67761f4fbc6fR87-R135) and [tests](https://github.com/stakwork/sphinx-meme/pull/10/files#diff-c22d481e1f8e0077946e1d2f2e86b84cd21b6749d7d9ca2989fb94bbd5c70fd1) that I have in the other project that leverages this functionality. 

In my opinion though, I do think that this is a _required_ type of functionality for something like aperture to support considering so many macaroon and lsat projects have it built in including [lncli](https://github.com/lightningnetwork/lnd/blob/28ea2736a04566b06c38533e5e48e1b8d185a49b/macaroons/constraints.go#L74-L81), [macaroon.js](https://www.npmjs.com/package/macaroons.js#verifying-macaroons-with-caveats), [boltwall](https://github.com/Tierion/boltwall/blob/master/src/configs/time.ts), and aperture [hints at it](https://github.com/lightninglabs/aperture/issues/80#issue-1380261086) with the sample config although the absolute (as opposed to relative) construction doesn't seem usable in many production use cases.

ellemouton:
Hey @bucko13 - sorry for the late reply, I have been OOO for a while.

>  As best as I can tell, there's no way that aperture can know if this volume has been surpassed or not and it will have to be checked by the service on the backend anyway. In fact, for aperture to have any sort of flexibility most of the validation beyond just restricting access to specific paths is going to have to be done after aperture has done a surface level validation of the macaroon.

Yes but the difference is that in that example, aperture doesnt know what the constraint is that it is adding but in this timeout case, it does. In this PR it is even classified in a different category to the other static constraints. with the other constraints aperture blindly adds them to the macaroon as caveats and it is the responsibility of the user to make sure that the backend server will know how to verify the caveat condition. But if we are making a whole new `Timeout` field - it is not clear that the user needs to make sure that the backend service should know how to verify that the macaroon has/has not timed out. So I think that if aperture is going to handle this timeout case specifically, then it should also verify the caveat. 

Thoughts?

(ps: I will try reply faster from now on 😅)

bucko13:
Hi @ellemouton, now it's my turn to apologize for the delayed response!

> But if we are making a whole new Timeout field - it is not clear that the user needs to make sure that the backend service should know how to verify that the macaroon has/has not timed out. So I think that if aperture is going to handle this timeout case specifically, then it should also verify the caveat.

Yep, I agree that aperture would be the best place for the verification. Since there was no other verification happening in aperture beyond the cryptographic macaroon verifications, I thought there might've been a reason for  leaving others out, including the `valid_until` in the sample conf. But if that's something you'd be ok accepting into the project, I'd be happy to move the validation code I [wrote in sphinx](https://github.com/stakwork/s
phinx-meme/pull/10/files#diff-28ef562dd4a2fb366dc11e80f6daa38220b2d132d2878c752d2f67761f4fbc6fR87-R135) over here. 

As for generic dynamic caveats, to just try and understand the proposal better, you're suggesting maybe NOT having the verification in aperture but rather having a way to tell aperture that you want to support a dynamic caveat that will be generated and provided by the backend service which would then be responsible for verifying them? 

So, something like this in the config.yml:

```yaml
dynamiccaveats:
    host: "127.0.0.1:8888"
    path: "/dynamic-caveats"
```

Then when the macaroon is baking, if this setting is on, aperture will call `GET 127.0.0.1:8888/dynamic-caveats` and get back some response that looks like:

```json
{
     "caveats": ["timeout=1670903314095", "challenge=e852397d-dd33-4808-b98e-2ee59b4cf5fb"] 
}
```

and then it can just be assumed that if a backend service is providing the caveats then it can validate them. (Note, I just came up with this off the top of my head. Definitely happy to workshop what the API looks like)

I think this seems reasonable to me. I'm trying to think of the best way to keep it nested and associated with a particular service, i.e. still support `service2_timeout=1670903314095` without making the API too complex.  

The only other downside of this is that it does significantly (imo) complicate the requirements for a backend service that needs/wants to leverage this. The nice thing about doing entirely via macaroons is that while the backend service does have to be able to validate the caveats, that's a pretty universal API. Knowing how to send and receive messages in a way aperture can understand is more complex than just having the information embedded in a standardized format in the Header. And then you also have to take into account compatibility, basically versioning the api the two servers use to talk to each other in case something changes. 

Really the best solution imo would be some kind of plugin and/or middleware system where rather than telling aperture "talk to this endpoint to figure out what other caveats to add", you would tell aperture: "load this set of middleware to add and verify custom caveats". So this makes it configurable and also keeps the verification and caveat generation in the same place. 

In fact, this is basically what I built for the Nodejs version of aperture, Boltwall. It's a little different since Boltwall is a middleware rather than a proxy server, but the idea is that when you add boltwall to your server, you can [pass in custom configs](https://github.com/Tierion/boltwall#custom-authorization-w-macaroons) when you initialize the middleware. One of the built-in configs that comes with boltwall is even a [timeout caveat](https://github.com/Tierion/boltwall#time_caveat_configs) exactly like what we're talking about!

Designing a plugin system for aperture though would obviously be a much more complex project to build.

ellemouton:
> Designing a plugin system for aperture though would obviously be a much more complex project to build.

completely agree. I think this would be a future feature :) I think `valid_until` is generic enough that aperture can just handle it. So yeah, if you want to move your validation code into aperture as a PR then that would be great!

bucko13:
Fantastic! My plate has been clearing up after a very busy few months so I’ll try and get an update up soon.

as an aside, a plug-in system would be REALLY cool and very appropriate for a system like this, but yeah, out of scope.

ellemouton:
> Fantastic! My plate has been clearing up after a very busy few months so I’ll try and get an update up soon.

Agreed!!!

> as an aside, a plug-in system would be REALLY cool and very appropriate for a system like this, but yeah, out of scope.

Completely agree!! Im hopeful that more people/groups will realise the value of LSATs soon which will give us more incentive to sink some time into cool things like that

bucko13:
@ellemouton take a look! the timeout caveat (which is set as service suffix + `_valid_until`) is now validated by aperture as well and there are tests for the satisfier and the minter's verifier.

ellemouton:
Awesome @bucko13 ! Before I jump into review, do you mind consolidating the commits a bit? for example, there are commits where you add things but then remove/change/format them in later commits. It will make reviewing a lot easier and is in any case what will be needed before merging :) Thanks!

bucko13:
@ellemouton no problem. Hope this is better!

bucko13:
Bumping this on request of the bot

mint/mint_test.go
```diff
@@ -225,3 +225,53 @@ func TestDemotedServicesLSAT(t *testing.T) {
...
+	}
+
+	// add a timeout caveat that expires in the future
+	timeout := lsat.NewTimeoutCaveat(testService.Name, 1000, &TestTime{})
+
+	err = lsat.AddFirstPartyCaveats(mac, timeout)
+	if err != nil {
+		t.Fatalf("unable to add caveat to LSAT: %v", err)
+	}
+
```
ellemouton, line 262: should call Verify LSAT here to check that the request is authorised if the timeout has not yet been reached

ellemouton (commented):
Very Cool PR & great test coverage!! 🔥 

Left quite a few comments and suggestions :)

Gonna be great to have this functionality!

bucko13:
Thanks for the review @ellemouton! will push up some updates asap. Quick question as a relatively new Golang developer- do you have any suggestions for setting up local formatting issues like the ones you commented on? I have the built in linter from VSCode, and used `go fmt` but obviously it didn't catch some rules for the code base.

lsat/service.go
```diff
@@ -129,3 +134,15 @@ func NewCapabilitiesCaveat(serviceName string, capabilities string) Caveat {
 		Value:     capabilities,
 	}
 }
+
+// NewTimeoutCaveat creates a new caveat for calculating an expiration 
+// for an lsat [seconds] from creation time.
+func NewTimeoutCaveat(serviceName string, seconds int64) Caveat {
+			macaroonTimeout := time.Duration(seconds)
+			requestTimeout := time.Now().Add(time.Second * macaroonTimeout).Unix()
```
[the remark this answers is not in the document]
bucko13: Ah ha. This is the `lll` is the linter I'm guessing?

guggero:
Unfortunately there is no good formatter (that I know of) that does this the same way `prettier` would for JS/TS automatically. It's the one thing the Golang ecosystem is missing IMO. We can tune the linter to be more pedantic about formatting things, but that doesn't solve the main issue of fixing it automatically, unfortunately.

ellemouton:
yeah - @bucko13 , I think the best would be to read over the [LND contribution guidelines](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md)
Also, seems like lots of the linter issues here could be solved by running `go fmt` in the same commit as the code is added in :)

lsat/service.go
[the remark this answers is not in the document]
ellemouton: oh yeah - it is just a way of saying "line length too long". ie, wrap at 80 chars :)

lsat/satisfier.go
```diff
@@ -115,3 +117,58 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
+			}
+
+			expirationTime := time.Unix(expirationTimestamp, 0)
+			currentTime := time.Unix(currentTimestamp, 0)
+			
+			if currentTime.Before(expirationTime) {
+				return nil
+			}
+			
+			return fmt.Errorf("not authorized to upload file. LSAT has expired")
```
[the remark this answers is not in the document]
bucko13: Whoops, sorry that's left over from the other application where I implemented this originally.

lsat/satisfier_test.go
```diff
@@ -0,0 +1,82 @@
...
+			final := tt.expirationValues[len(tt.expirationValues)-1]
+			finalCaveat := NewCaveat(condition, final)
+			err := satisfier.SatisfyFinal(finalCaveat)
+
+			if err != nil && tt.successFinal {
+				t.Fatal("unexpectedly failed on SatisfyFinal")
+			}
+
+			if err == nil && !tt.successFinal {
+				t.Fatalf("unexpectedly passed on SatisfyFinal")
```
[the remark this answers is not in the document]
bucko13: Would that be like this:

```
require.NoError(t, "unexpectedly passed on SatisfyFinal")
```

bucko13:
> Also, seems like lots of the linter issues here could be solved by running go fmt in the same commit as the code is added in :)

Hmmm I'm pretty sure I did that but maybe not. Since all of this is basically just one feature, would you have any objection to my just applying `go fmt`, addressing all the remaining issues, and then just squashing into a single commit? 

(sorry got pulled off in other directions again. I'm determined to wrap this up this week though if possible and will find the time assuming nothing really crazy comes up in the meantime)

ellemouton:
HI @bucko13 - yeah I think the diff is small enough & all the parts fit together quite tightly so having them in one commit should be fine :)

bucko13:
For autoformatting with vscode I actually found the combination of [golines](https://github.com/segmentio/golines) plus the [runonsave extension](https://github.com/emeraldwalk/vscode-runonsave) to work pretty well in helping me clean things up a little easier.

lsat/satisfier.go
```diff
@@ -115,3 +117,58 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
 	}
 }
+
+// A satisfier to check if an LSAT is expired or not.
+// The Satisfier takes a service name to set as the condition prefix 
+// and currentTimestamp to compare against the expiration(s) in the caveats. The the expiration time
+// is retrieved from the caveat values themselves. The satisfier
+// will also make sure that each subsequent caveat of the same condition only has increasingly
+// strict expirations.
+func NewTimeoutSatisfier(service string, currentTimestamp int64) Satisfier {
```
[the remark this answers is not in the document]
bucko13: I like the change for the caveat code you suggested along these lines, but I think it feels cleaner to keep the satisfier just taking the value directly. Imo it makes the tests a bit cleaner as well as for the actual comparison with the expiration time which is coming in directly as an int64 from the caveat value, making [these two lines](https://github.com/lightninglabs/aperture/pull/81/files?diff=unified&w=0#diff-28ef562dd4a2fb366dc11e80f6daa38220b2d132d2878c752d2f67761f4fbc6fR164-R165) have nice and clean parallels

```go
expirationTime := time.Unix(expirationTimestamp, 0)
currentTime := time.Unix(currentTimestamp, 0)
```

lsat/satisfier_test.go
```diff
@@ -0,0 +1,82 @@
...
+				if err != nil && tt.successPrev {
+					t.Fatalf("unexpectedly failed on SatisfyPrevious comparing prev %s to current %s", prev, final)
+				}
+
+				if err == nil && !tt.successPrev {
+					t.Fatalf("unexpectedly passed on SatisfyPrevious comparing prev %s to current %s", prev, final)
+				}
+
+			}
+		})
```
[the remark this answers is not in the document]
bucko13: I like it! Thanks.

bucko13:
Thanks for the thorough review @ellemouton! I think all comments have been addressed (and all commits squashed into one).

services.go
```diff
@@ -22,16 +24,28 @@ var _ mint.ServiceLimiter = (*staticServiceLimiter)(nil)
 
 // newStaticServiceLimiter instantiates a new static service limiter backed by
 // the given restrictions.
-func newStaticServiceLimiter(proxyServices []*proxy.Service) *staticServiceLimiter {
+func newStaticServiceLimiter(
+	proxyServices []*proxy.Service,
+) *staticServiceLimiter {
```
guggero, line 29: I tried out `golines`, but unfortunately it doesn't produce the function signatures the way we want them. This should be formatted as:

```
func newStaticServiceLimiter(
	proxyServices []*proxy.Service) *staticServiceLimiter {
```

instead. Too bad there doesn't seem to be a formatting tool out there that we can configure that way...

lsat/satisfier.go
```diff
@@ -115,3 +120,63 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
+// and currentTimestamp to compare against the expiration(s) in the caveats.
+// The expiration time is retrieved from the caveat values themselves.
+// The satisfier will also make sure that each subsequent caveat of the same
+// condition only has increasingly strict expirations.
+func NewTimeoutSatisfier(service string, currentTimestamp int64) Satisfier {
+	return Satisfier{
+		Condition: service + CondTimeoutSuffix,
+
+		SatisfyPrevious: func(prev, cur Caveat) error {
+			// check that each caveat is more restrictive than the previous
```
ellemouton, line 135: that check is not happening here though but a few lines down. Also, start of sentence should start with a capital letter.

lsat/satisfier.go
```diff
@@ -115,3 +120,63 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
+			prevValue, prevValErr := strconv.ParseInt(prev.Value, 10, 64)
+			currValue, currValErr := strconv.ParseInt(cur.Value, 10, 64)
+
+			if prevValErr != nil {
+				return fmt.Errorf("caveat value not a valid integer: %v", prevValErr)
+			}
+
+			if currValErr != nil {
+				return fmt.Errorf("caveat value not a valid integer: %v", currValErr)
```
ellemouton, line 144: errors should be checked and handled where the are created:

```
	                prevValue, err := strconv.ParseInt(prev.Value, 10, 64)
			if err != nil {
				return fmt.Errorf("error parsing previous "+
					"caveat value: %w", err)
			}

			currValue, err := strconv.ParseInt(cur.Value, 10, 64)
			if err != nil {
				return fmt.Errorf("error parsing caveat "+
					"value: %w", err)
			}
```

lsat/satisfier.go
```diff
@@ -115,3 +120,63 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
+// NewTimeoutSatisfier checks if an LSAT is expired or not.
+// The Satisfier takes a service name to set as the condition prefix
+// and currentTimestamp to compare against the expiration(s) in the caveats.
+// The expiration time is retrieved from the caveat values themselves.
+// The satisfier will also make sure that each subsequent caveat of the same
+// condition only has increasingly strict expirations.
+func NewTimeoutSatisfier(service string, currentTimestamp int64) Satisfier {
+	return Satisfier{
+		Condition: service + CondTimeoutSuffix,
+
```
ellemouton, line 133: nit: remove new line

lsat/satisfier.go
```diff
@@ -115,3 +120,63 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
+			// earlier than ones after it b/c that means they are getting
+			// more permissive.
+			if prevTime.Before(currTime) {
+				return fmt.Errorf("%s caveat violates increasing "+
+					"restrictiveness", service+CondTimeoutSuffix)
+			}
+
+			return nil
+		},
+
```
ellemouton, line 160: nit: remove new line

lsat/satisfier.go
```diff
@@ -115,3 +120,63 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
+		SatisfyFinal: func(c Caveat) error {
+			// make sure that the final relevant caveat is not passed
+			// the current date/time
+			expirationTimestamp, err := strconv.ParseInt(c.Value, 10, 64)
+
```
ellemouton, line 165: nit: remove new line and wrap correctly

lsat/satisfier.go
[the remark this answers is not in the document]
ellemouton: but if you use `now func() time.Time` then you can do:

```
expirationTime := time.Unix(expirationTimestamp, 0)

if now().Before(expirationTime) {
		return nil
}
```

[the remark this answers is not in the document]
ellemouton: meaning you only have to convert 1 value instead of 2.

lsat/service.go
```diff
@@ -15,6 +16,10 @@ const (
 	// capabilities caveat. For example, the condition of a capabilities
 	// caveat for a service named `loop` would be `loop_capabilities`.
 	CondCapabilitiesSuffix = "_capabilities"
+
+	// CondTimeoutSuffix is the condition suffix used for a service's timeout
+	// caveat.
```
ellemouton, line 21: not wrapped at 80

lsat/service.go
```diff
@@ -129,3 +134,18 @@ func NewCapabilitiesCaveat(serviceName string, capabilities string) Caveat {
 		Value:     capabilities,
 	}
 }
+
+// NewTimeoutCaveat creates a new caveat for calculating an expiration
+// for an lsat [seconds] from creation time.
+func NewTimeoutCaveat(serviceName string, seconds int64,
```
ellemouton, line 140: suggest:

```
// NewTimeoutCaveat creates a new caveat that will result in a macaroon being
// valid for numSeconds after the current time.
func NewTimeoutCaveat(serviceName string, numSeconds int64,
	now func() time.Time) Caveat {

	var (
		macaroonTimeout = time.Duration(numSeconds) * time.Second
		requestTimeout  = now().Add(macaroonTimeout)
	)

	return Caveat{
		Condition: serviceName + CondTimeoutSuffix,
		Value:     strconv.FormatInt(requestTimeout.Unix(), 10),
	}
}
```

mint/interface.go
```diff
@@ -0,0 +1,14 @@
+package mint
+
+import (
+	"time"
+)
+
+// TimeProvider is a way to get a Now() method while
+// also being able to mock the same for tests
+type TimeProvider interface {
```
ellemouton, line 9: dont think we need this :)

mint/mint.go
```diff
@@ -60,6 +61,10 @@ type ServiceLimiter interface {
 	// enforces additional constraints on a particular service/service
 	// capability.
 	ServiceConstraints(context.Context, ...lsat.Service) ([]lsat.Caveat, error)
+
+	// ServiceTimeouts returns the timeout caveat for each service. This will
+	// determine if and when service access can expire
```
ellemouton, line 66: pls wrap at 80 🙏

mint/mint_test.go
```diff
@@ -225,3 +225,58 @@ func TestDemotedServicesLSAT(t *testing.T) {
 		t.Fatal("expected macaroon to be invalid")
 	}
 }
+
+func TestExpiredServicesLSAT(t *testing.T) {
```
ellemouton, line 229: missing godoc

mint/mint.go
```diff
@@ -78,15 +83,26 @@ type Config struct {
...
+	now := time.Now()
+	return now
+}
+
 // Mint is an entity that is able to mint and verify LSATs for a set of
 // services.
 type Mint struct {
 	cfg Config
+
+	timeProvider TimeProvider
```
ellemouton, line 100: should be in Config. 

Suggest:
```
type Config struct {
    ....
    // Now returns the current time.
    Now func() time.Time
}
```

mint/mint_test.go
```diff
@@ -225,3 +225,58 @@ func TestDemotedServicesLSAT(t *testing.T) {
...
+		Challenger:     newMockChallenger(),
+		ServiceLimiter: newMockServiceLimiter(),
+	}, &TestTime{})
+
+	// Mint a new lsat for accessing a test service
+	mac, _, err := mint.MintLSAT(ctx, testService)
+	if err != nil {
+		t.Fatalf("unable to mint LSAT: %v", err)
+	}
+
```
ellemouton, line 244: pls use the `require` package rather

proxy/service.go
```diff
@@ -72,6 +72,12 @@ type Service struct {
 	// the file is sent encoded as base64.
 	Headers map[string]string `long:"headers" description:"Header fields to always pass to the service"`
 
+	// Timeout is an optional value that indicates in how many seconds the
+	// services caveat should timeout relative to the time of creation. So
```
ellemouton, line 76: s/services/service's

ellemouton: s/timeout/time out

proxy/service.go
```diff
@@ -72,6 +72,12 @@ type Service struct {
...
+	// if a value of 100 is set, then the timeout will be 100 seconds
+	// after creation.
```
ellemouton, line 78: "after creation of the LSAT"

proxy/service.go
```diff
@@ -72,6 +72,12 @@ type Service struct {
...
+	Timeout int64 `long:"timeout" description:"An integer value that indicates how many seconds until the service access expires"`
```
ellemouton, line 79: suggest "the number of seconds" instead of "how many seconds"

ellemouton (commented):
Thanks for the updates @bucko13 !

See my attached patch that you can use to apply most of the suggestions I have: 

[timeout.patch](https://github.com/lightninglabs/aperture/files/11100454/timeout.patch)

lsat/satisfier.go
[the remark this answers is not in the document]
bucko13: Ah ok. Didn't realize go's time had comparisons like that. Very nice.

services.go
bucko13: Gotcha. Yeah, tbh this is an issue with most formatters regardless of language in my experience. I guess it really comes down to a tradeoff between desired style vs. automated and reproducible standards. Line lengths are notoriously difficult to get "good" results (I imagine that's why `gofmt` left them off entirely).

lsat/interface.go
```diff
@@ -0,0 +1,14 @@
...
+)
+
+// TimeProvider is a way to get a Now() method while
+// also being able to mock the same for tests
+type TimeProvider interface {
+	// In most cases Now() should just return the result
+	// of Go's time.Now method, but sometimes, e.g. for tests
+	// it can be useful to have a mocked version of this.
+	Now() time.Time
+}
```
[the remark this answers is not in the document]
bucko13: Sorry, I think I misunderstood your comment originally. Should be resolved now with your patch (removed entirely now)

bucko13:
Thank you for your patch and your patience elle! Very nice cleanup. Hopefully everything's been addressed now.

aperture.go
```diff
@@ -627,7 +627,7 @@ func createProxy(cfg *Config, challenger *LndChallenger,
 		Challenger:     challenger,
 		Secrets:        newSecretStore(etcdClient),
 		ServiceLimiter: newStaticServiceLimiter(cfg.Services),
-	})
+	}, &mint.GetTime{})
```
ellemouton, line 630: not required anymore - also results in a compile error

lsat/satisfier.go
```diff
@@ -115,3 +119,64 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
 		},
 	}
 }
+
+// NewTimeoutSatisfier checks if an LSAT is expired or not. The Satisfier takes
```
ellemouton, line 123: previous comment resolved but not addressed yet - pls wrap things at 80 chars :)

lsat/satisfier.go
```diff
@@ -115,3 +119,64 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
+			expirationTime := time.Unix(expirationTimestamp, 0)
+
+			// Make sure that the final relevant caveat is not
+			// passed the current date/time
+			if now().Before(expirationTime) {
+				return nil
+			}
+
+			return fmt.Errorf(
+				"not authorized to access service. LSAT has " +
```
ellemouton, line 177: nit: since this is a returned error, this should be formatted like this:

```
	return fmt.Errorf("not authorized to access " +
				"service. LSAT has expired")
```

mint/mint.go
```diff
@@ -7,10 +7,10 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
-
 	"github.com/lightninglabs/aperture/lsat"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"gopkg.in/macaroon.v2"
+	"time"
```
ellemouton, line 13: imports should be: first standard lib imports, new line, other imports. So `time` belongs in the standard section and the new line should stay

mint/mint_test.go
```diff
@@ -5,6 +5,9 @@ import (
 	"crypto/sha256"
 	"strings"
 	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
```
ellemouton, line 10: nit: remove new line below

mint/mock_test.go
```diff
@@ -3,10 +3,9 @@ package mint
 import (
 	"context"
 	"crypto/sha256"
-	"math/rand"
-
 	"github.com/lightninglabs/aperture/lsat"
 	"github.com/lightningnetwork/lnd/lntypes"
+	"math/rand"
```
ellemouton, line 8: same comment re imports: `math/rand` belongs in the standard lib section and new line should stay

proxy/service.go
```diff
@@ -72,6 +72,12 @@ type Service struct {
 	// the file is sent encoded as base64.
 	Headers map[string]string `long:"headers" description:"Header fields to always pass to the service"`
 
+	// Timeout is an optional value that indicates in how many seconds the
+	// services caveat should timeout relative to the time of creation. So
+	// if a value of 100 is set, then the timeout will be 100 seconds
```
ellemouton, line 77: comment resolved but hasnt been fixed yet

proxy/service.go
```diff
@@ -72,6 +72,12 @@ type Service struct {
...
+	// after creation.
+	Timeout int64 `long:"timeout" description:"An integer value that indicates the number of seconds until the service access expires"`
```
ellemouton, line 79: same here: comment resolved but not fixed yet

ellemouton (commented):
Getting there!

bucko13:
Ok, should be addressed now. There are fewer remaining comments so it was easier to keep track without marking as resolved so I left them all unresolved to avoid any other mixups. Also rebased with master.

ellemouton:
Thanks for the fast updates @bucko13! 🚀  Soooo close - just checkout the remaining linter error: basically I think you just need to add a `test := test` line in the satisfier test for loop that iterates over the tests (it's a weird go thing)

Yeah - sorry about the line length things - the older code in the repo doesnt match the standard that we try stick to now 🤷‍♀️  I know it can be annoying - but thanks for your patience & fast updates :)

bucko13:
Nice, found [this](https://stackoverflow.com/questions/68559574/using-the-variable-on-range-scope-x-in-function-literal-scopelint) on that linting error.

bucko13:
Yay, green checks!

ellemouton (approved):
Excellent work @bucko13!! tACK! 🔥 

The last thing that should be added is a `timeout` entry in the `sample-conf.yaml` file.

Once that is added, then this LGTM! 🚀

lsat/satisfier.go
```diff
@@ -115,3 +119,62 @@ func NewCapabilitiesSatisfier(service string, targetCapability string) Satisfier
...
+			)
+			if err != nil {
+				return fmt.Errorf("caveat value not a valid "+
+					"integer: %v", err)
+			}
+
+			expirationTime := time.Unix(expirationTimestamp, 0)
+
+			// Make sure that the final relevant caveat is not
+			// passed the current date/time
```
guggero, line 171: nit: missing full stop.

mint/mint.go
```diff
@@ -60,6 +61,10 @@ type ServiceLimiter interface {
 	// enforces additional constraints on a particular service/service
 	// capability.
 	ServiceConstraints(context.Context, ...lsat.Service) ([]lsat.Caveat, error)
+
+	// ServiceTimeouts returns the timeout caveat for each service. This
+	// will determine if and when service access can expire
```
guggero, line 66: nit: missing full stop.

guggero (approved):
Nice feature, thank you! LGTM pending @ellemouton's last comment and my nits.

mint/mint.go
bucko13: Added.

lsat/satisfier.go
bucko13: added

bucko13:
Thanks @ellemouton!

> The last thing that should be added is a timeout entry in the sample-conf.yaml file.

Added this. One question I have is that there's currently an example `constraint` value added in the sample that is kind of made obsolete from this:

```
    # The set of constraints that are applied to tokens of the service at the
    # base tier.
    constraints:
        "valid_until": "2020-01-01"
```

This was obviously just for example purposes anyway so it's not a huge deal, but it is something y'all are sharing for an example so I didn't want to remove. Happy to change it to something else if there was an example from your own internal usage you thought would be better, or I could just leave it.

ellemouton:
good catch - I think we can leave it for now since it is not exactly the same. The "valid_until" is an absolute timeout whereas the new "timeout" is relative. I think there could be situations where a user would want both. But on the other hand, yes it is just an example so not such a big deal to change it.

guggero:
The timeout is created by specifying a relative value, yes. But for things to work it still needs to be an absolute timestamp. So I guess the only difference between the example `"valid_until": "2020-01-01"` and the implemented `_valid_until` suffix is the format of the timestamp, which would now be a Unix timestamp. 
So I do think we should just update the example to include a Unix timestamp instead of the yyyy-mm-dd example value and call it a day.

bucko13:
> So I do think we should just update the example to include a Unix timestamp instead of the yyyy-mm-dd example value and call it a day.

ok so you’re saying to update the valid_until example to be a timestamp as well but just for the corresponding date? To be clear, as far as I know it’s just an example right and doesn’t actually do anything in aperture today?

guggero:
> ok so you’re saying to update the valid_until example to be a timestamp as well but just for the corresponding date? To be clear, as far as I know it’s just an example right and doesn’t actually do anything in aperture today?

Correct. That was just an example of how Aperture could be extended with functionality, which this PR now did. Feel free to use whatever example timestamp you like.

bucko13:
> So I do think we should just update the example to include a Unix timestamp instead of the yyyy-mm-dd example value and call it a day.

Added!

PR #82, opened 2022-09-26
multi: gc stale mailboxes

ellemouton:
In this commit, we start a timer if a mailbox stream is completely 
un-occupied (neither read or write stream is occupied). The timer 
is stopped if either of the streams are occupied and is reset if 
both streams are unoccupied. If the timer ever ticks then the 
mailbox stream is torn down. The default "stale time" is set to one 
hour but this can be set via the `hashmail.staletimeout` flag.

### DIFF
--- a/hashmail_server.go
+++ b/hashmail_server.go
@@ -28,6 +28,10 @@ const (
 	// then we'll allow it up to this burst allowance.
 	DefaultMsgBurstAllowance = 10
 
+	// DefaultStaleTimeout is the time after which a mailbox will be torn
+	// down if neither of its streams are occupied.
+	DefaultStaleTimeout = time.Hour
+
 	// DefaultBufSize is the default number of bytes that are read in a
 	// single operation.
 	DefaultBufSize = 4096
@@ -185,11 +189,14 @@ type stream struct {
 	wg sync.WaitGroup
 
 	limiter *rate.Limiter
+
+	status *streamStatus
 }
 
 // newStream creates a new stream independent of any given stream ID.
 func newStream(id streamID, limiter *rate.Limiter,
-	equivAuth func(auth *hashmailrpc.CipherBoxAuth) error) *stream {
+	equivAuth func(auth *hashmailrpc.CipherBoxAuth) error,
+	onStale func() error, staleTimeout time.Duration) *stream {
 
 	// Our stream is actually just a plain io.Pipe. This allows us to avoid
 	// having to do things like rate limiting, etc as we can limit the
@@ -204,6 +211,7 @@ func newStream(id streamID, limiter *rate.Limiter,
 		id:              id,
 		equivAuth:       equivAuth,
 		limiter:         limiter,
+		status:          newStreamStatus(onStale, staleTimeout),
 		readBytesChan:   make(chan []byte),
 		readErrChan:     make(chan error, 1),
 		quit:            make(chan struct{}),
@@ -213,6 +221,7 @@ func newStream(id streamID, limiter *rate.Limiter,
 	// will cause the goroutine below to get an EOF error when reading,
 	// which will cause it to close the other ends of the pipe.
 	s.tearDown = func() error {
+		s.status.stop()
 		err := writeWritePipe.Close()
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


ellemouton:
cc @guggero

hashmail_server.go
```diff
@@ -198,12 +205,27 @@ func newStream(id streamID, limiter *rate.Limiter,
 	writeReadPipe, writeWritePipe := io.Pipe()
 	readReadPipe, readWritePipe := io.Pipe()
 
+	staleTimer := time.AfterFunc(staleTimeout, func() {
+		if err := onStale(); err != nil {
+			log.Errorf("error in onStale callback: %v", err)
+		}
+	})
+	defer staleTimer.Stop()
```
guggero, line 213: Can't use a `defer` here, as the `newStream` function will return immediately.

hashmail_server.go
```diff
@@ -710,3 +764,46 @@ func (h *hashMailServer) RecvStream(desc *hashmailrpc.CipherBoxDesc,
...
+
+// newStreamStatus constructs a new streamStatus instance.
+func newStreamStatus(onOccupied, onIdle func()) *streamStatus {
+	return &streamStatus{
+		onIdle:     onIdle,
+		onOccupied: onOccupied,
+	}
+}
+
+// streamStatus should be called when one of the sub-streams (read or write)
```
guggero, line 788: nit: Godoc comment incorrect.

hashmail_server.go
```diff
@@ -198,12 +205,27 @@ func newStream(id streamID, limiter *rate.Limiter,
 	writeReadPipe, writeWritePipe := io.Pipe()
 	readReadPipe, readWritePipe := io.Pipe()
 
+	staleTimer := time.AfterFunc(staleTimeout, func() {
```
guggero, line 208: Should we allow to disable this feature, for example by setting a value of `-1`? Just in case we notice it's breaking something and we need a quick fix.

hashmail_server.go
```diff
@@ -284,12 +306,14 @@ func newStream(id streamID, limiter *rate.Limiter,
 // ReturnReadStream returns the target read stream back to its holding channel.
 func (s *stream) ReturnReadStream(r *readStream) {
 	s.readStreamChan <- r
+	s.status.streamReturned()
```
guggero, line 309: Can we add a simple unit test to make sure these are counted at the correct moment? I'm not sure when those channels actually are sent to. Just to avoid surprises.

hashmail_server.go
guggero: I think we should either move the timer into the `streamStatus` and make it less generic or add the `staleTimer.Stop()` to the `tearDown` func.

guggero (commented):
Very nice feature!
Some unit tests would be helpful here though.

hashmail_server.go
ellemouton: > Should we allow to disable this feature, for example by setting a value of `-1`? Just in case we notice it's breaking something and we need a quick fix.

Great idea! Added 👍

hashmail_server.go
ellemouton: good idea - I added a test, let me know if this is what you had in mind :)

ellemouton:
Thanks @guggero ! I have updated it and have added a unit test. I hope I havent over-complicated it now... let me know what you think :)  (changed some things like using bools for the read&write stream occupancy instead of a counter)

hashmail_server_test.go
```diff
@@ -242,3 +245,263 @@ func readMsgFromStream(t *testing.T,
...
+			require.NoError(t, err)
+
+			hm.assertStreamsOccupied(test.readerConnected)
+
+			// Ensure that attempting to take the read stream and
+			// receive from it while it is currently occupied will
+			// result in an error.
+			err = recvFromStream(client2)
+			require.Error(t, err)
+			require.True(t, strings.Contains(err.Error(), "read "+
```
guggero, line 350: nit: `assert.Contains()`

hashmail_server_test.go
```diff
@@ -242,3 +245,263 @@ func readMsgFromStream(t *testing.T,
...
+
+			client1 := hashmailrpc.NewHashMailClient(conn1)
+			client2 := hashmailrpc.NewHashMailClient(conn2)
+
+			// Let client 1 create a mailbox on the server.
+			resp, err := client1.NewCipherBox(
+				ctx, &hashmailrpc.CipherBoxAuth{
+					Auth: &hashmailrpc.CipherBoxAuth_LndAuth{},
+					Desc: testStreamDesc,
+				})
```
guggero, line 321: nit: closing parenthesis on new line.

hashmail_server_test.go
```diff
@@ -242,3 +245,263 @@ func readMsgFromStream(t *testing.T,
...
+				readOccupied:  false,
+			},
+			senderDisconnected: statusState{
+				writeOccupied: false,
+				readOccupied:  false,
+			},
+		},
+	}
+
+	for _, test := range tests {
```
guggero, line 299: Nice test :100:

guggero (approved):
Very nice, exactly what I had in mind!
LGTM :tada:

PR #84, opened 2023-01-23
Upgrade lnd tor to version v1.1.0

ffranr:
The goal: update lnd/tor to version v1.1.0 so that we can use aperture in taro.


guggero:
Need to bump the Golang version in CI, the one being used now is pretty ancient.

tools/tools.go
```diff
@@ -0,0 +1,18 @@
+//go:build tools
+// +build tools
+
+package tarod
```
guggero, line 4: nit: need to change this to `aperture`.

tools/tools.go
```diff
@@ -0,0 +1,18 @@
...
+
+// The other imports represent our build tools. Instead of defining a commit we
+// want to use for those golang based tools, we use the go mod versioning system
+// to unify the way we manage dependencies. So we define our build tool
+// dependencies here and pin the version in go.mod.
+import (
+	_ "github.com/btcsuite/btcd"
+	_ "github.com/dvyukov/go-fuzz/go-fuzz"
```
guggero, line 12: I think we only need `golangci-lint`, `go-acc` and `gosimports` here, should clean up the go.mod somewhat.

tools/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/lightninglabs/taro/tools
```
guggero, line 1: This needs to be `github.com/lightninglabs/aperture/tools`.

onion_store.go
```diff
@@ -44,21 +44,20 @@ func onionPath(onionType tor.OnionType) (string, error) {
 // onionStore is an etcd-based implementation of tor.OnionStore.
 type onionStore struct {
 	*clientv3.Client
+	OnionType tor.OnionType
```
guggero, line 47: nit: un-export since the struct itself is not exported either? Also I'd add a newline between embedded and non-embedded struct members.

guggero: The alternative would be to remove V2 support completely as that's not supported by the Tor network anymore anyway (since October last year I think).

aperture.go
```diff
@@ -612,37 +612,34 @@ func getTLSConfig(serverName, baseDir string, autoCert bool) (
 // specified in the config. Onion services will be created over which the proxy
 // can be reached at.
 func initTorListener(cfg *Config, etcd *clientv3.Client) (*tor.Controller, error) {
+	// Determine tor onion type based on config flags.
+	var onionType tor.OnionType
+	if cfg.Tor.V2 {
```
guggero, line 617: See my previous comment about V2 being deprecated. Would clean up the config and code quite a bit if we removed it.

tools/go.mod
```diff
@@ -0,0 +1,13 @@
...
+
+go 1.16
```
guggero, line 3: Use 1.19 here as well?

hashmail_server_test.go
```diff
@@ -86,10 +86,10 @@ func TestHashMailServerReturnStream(t *testing.T) {
 
 	// Send then receive yet another message to make sure the stream is
 	// still operational.
-	testMessage2 := append(testMessage, []byte("test")...)
+	testMessage := append(testMessage, []byte("test")...)
```
guggero, line 89: Hmm, this is weird, since it _does_ create a new variable (with the same name but in a new scope). I'd say this might be a candidate for a `//nolint` directive instead, as this looks more like a headscratcher than before IMO.

guggero (commented):
Very nice, great work on this!

I think all the changes to the mocks probably need to be squashed into the same commit as the `lnd` dependency bump, as otherwise that commit won't compile (we don't have a check for this in the CI in this repo yet).

onion_store.go
ffranr: I've opted to remove V2 support completely.

aperture.go
ffranr: Will remove tor v2 support, following a suggestion from your other comment.

guggero (approved):

.editorconfig
```diff
@@ -0,0 +1,16 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
ellemouton, line 1: oh cool!

onion_store.go
```diff
@@ -2,7 +2,6 @@ package aperture
 
 import (
 	"context"
```
ellemouton, line 4: nit: commit title should ideally be no more than 50 chars and commit messages should be wrapped at 72 since `git log` doesn't do any formatting: https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#model-git-commit-messages

go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/lightninglabs/aperture
 go 1.15
```
ellemouton, line 3: i think we should bump this go version too?

tools/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.19.2-buster
```
ellemouton, line 1: nit: same thing re commit message

ellemouton (commented):
So glad that this is being taken care of!! This will save me a lot of headache on the Litd&LNC side 🙌 

Looks good - just a few notes. Main thing is that I think the main go version should also be bumped

go.mod
ffranr: Good point, might as well!

ellemouton (approved):
Noice 🔥

ffranr:
@guggero @ellemouton Thanks for the reviewing!

I don't have write access to this repository. Can either of you merge please?

PR #91, opened 2023-03-06
multi: create challenger, secret, and interceptor packages

orbitalturtle:
This PR is just moving things around and exporting things. It attempts to make a few organizational changes:
- Moves the challenger, secret, and client_interceptor code to new packages so that 1) they can be imported elsewhere. 2) to get rid of importing cycles when importing this code (particularly for the watchtower case described below)
- Exports some of the testing mock tools, since they are potentially useful in the watchtower tests, and seems to make sense to keep things as DRY as possible!

Stepping back, the higher level goal here is to be able to use some of the minter and challenger code in the lnd watchtower for minting and verifying lsats. I had talked with @roasbeef a while back about adding a method of paying a watchtower for storing state updates for other users, and we discussed using lsats for this purpose. This lays the groundwork for that. (Initial PR for the watchtower code incoming soon(tm))

If there are reasons to organize these changes in a different way, I'm more than happy to make adjustments, thanks!


ellemouton:
Thanks for the PR @orbitalturtle 🔥 

@guggero  - I dont have the perms to request myself as reviewer here it seems. Can you request review from me pls?

bucko13:
Just wanted to chime in a plus one on the motivation for this. Being able to use this more as a library when applicable will be huge for lsat adoption imo.

secrets/secrets.go
```diff
@@ -40,9 +50,9 @@ type secretStore struct {
 // A compile-time constraint to ensure secretStore implements mint.SecretStore.
 var _ mint.SecretStore = (*secretStore)(nil)
 
-// newSecretStore instantiates a new LSAT secrets store backed by an etcd
+// NewSecretStore instantiates a new LSAT secrets store backed by an etcd
 // cluster.
-func newSecretStore(client *clientv3.Client) *secretStore {
+func NewSecretStore(client *clientv3.Client) *secretStore {
```
ellemouton, line 55: to avoid stuttering (`secretes.NewSecretStore(...)`), how about just `NewStore`?

aperture.go
```diff
@@ -18,9 +18,12 @@ import (
 	gateway "github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	flags "github.com/jessevdk/go-flags"
 	"github.com/lightninglabs/aperture/auth"
+	"github.com/lightninglabs/aperture/challenger"
```
ellemouton, line 21: commit message comment:

- `NewLndChallenger` isnt altered to take in an lndclient (which is the thing that would cause an import cycle), but rather is altered to take a `challenger.InvoiceClient` (which lnd client just happens to implement)
- nitty nit: check out LND contribution guidelines for ideal commit title & message structure 🙏

ellemouton: also - for easier review, I think perhaps do 1 commit where you do the refactoring and then a separate commit for the `NewLndChallenger` change

log.go
```diff
@@ -27,6 +28,7 @@ func SetupLoggers(root *build.RotatingLogWriter, intercept signal.Interceptor) {
 
 	lnd.SetSubLogger(root, Subsystem, log)
 	lnd.AddSubLogger(root, auth.Subsystem, intercept, auth.UseLogger)
+	lnd.AddSubLogger(root, challenger.Subsystem, intercept, challenger.UseLogger)
```
ellemouton, line 32: nit: wrap at 80 🙏

challenger/challenger.go
```diff
@@ -63,26 +62,18 @@ var _ auth.InvoiceChecker = (*LndChallenger)(nil)
...
-	invoiceMacaroonName = "invoice.macaroon"
+	InvoiceMacaroonName = "invoice.macaroon"
 )
 
 // NewLndChallenger creates a new challenger that uses the given connection
 // details to connect to an lnd backend to create payment challenges.
-func NewLndChallenger(cfg *AuthConfig, genInvoiceReq InvoiceRequestGenerator,
-	errChan chan<- error) (*LndChallenger, error) {
+func NewLndChallenger(genInvoiceReq InvoiceRequestGenerator,
+	errChan chan<- error, client InvoiceClient) (*LndChallenger, error) {
```
ellemouton, line 71: micro nit: I think keep the errChan as the last parameter.

secrets/test_util.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/server/v3/embed"
+)
+
+// EtcdSetup is a helper that instantiates a new etcd cluster along with a
+// client connection to it. A cleanup closure is also returned to free any
+// allocated resources required by etcd.
+func EtcdSetup(t *testing.T) (*clientv3.Client, func()) {
```
ellemouton, line 17: can remove the `etcdSetup` func that is still in the `secrets_test.go` file since it is no longer used

log.go
ellemouton: same for `interceptor.Subsystem` line

interceptor/log.go
```diff
@@ -0,0 +1,26 @@
+package interceptor
+
+import (
+	"github.com/btcsuite/btclog"
+	"github.com/lightningnetwork/lnd/build"
+)
+
+// Subsystem defines the sub system name of this package.
+const Subsystem = "INPTR"
```
ellemouton, line 9: nit: usually we keep these to 4 chars

interceptor/client_interceptor.go
```diff
@@ -88,7 +89,7 @@ type ClientInterceptor struct {
 // NewInterceptor creates a new gRPC client interceptor that uses the provided
 // lnd connection to automatically acquire and pay for LSAT tokens, unless the
 // indicated store already contains a usable token.
-func NewInterceptor(lnd *lndclient.LndServices, store Store,
+func NewInterceptor(lnd *lndclient.LndServices, store lsat.Store,
```
ellemouton, line 92: how about just `New`?

ellemouton: or `NewClientInterceptor`

interceptor/client_interceptor.go
```diff
@@ -407,7 +408,7 @@ func (i *ClientInterceptor) payLsatToken(ctx context.Context, md *metadata.MD) (
 
 // trackPayment tries to resume a pending payment by tracking its state and
 // waiting for a conclusive result.
-func (i *ClientInterceptor) trackPayment(ctx context.Context, token *Token) error {
+func (i *ClientInterceptor) trackPayment(ctx context.Context, token *lsat.Token) error {
```
ellemouton, line 411: nit: line too long

ellemouton (commented):
Looking good!

orbitalturtle:
@ellemouton Yikes, yes, sorry for being so slow :sloth: to get to this. I don't know how time moves so fast. I'll review and address your feedback this weekend. :pray:

aperture.go
orbitalturtle: I moved the NewLndChallenger changes into a new commit. The change was a bit awkward because NewLndChallenger initially needs AuthConfig, which is defined in the main package. But then in the next commit we remove AuthConfig entirely because it's not needed once we update NewLndChallenger to take in the client as a parameter. This was the cleanest way I could think to make the change - maybe it's still worth it for review readability. Lmk your thoughts!

orbitalturtle:
@ellemouton  Finally got around to this. Sorry for the long wait. Promise if you have any more feedback I'll respond much faster! One comment/question above but otherwise was able to address the rest of the comments.

Roasbeef:
Approved the CI run.

orbitalturtle:
Thanks @Roasbeef - fixed that linter issue.

auth/test_utils.go
```diff
@@ -0,0 +1,61 @@
...
+// CreateDummyMacHex creates a valid macaroon with dummy content for our tests.
+func CreateDummyMacHex(preimage string) string {
+	dummyMac, err := macaroon.New(
+		[]byte("aabbccddeeff00112233445566778899"), []byte("AA=="),
+		"aperture", macaroon.LatestVersion,
+	)
+	if err != nil {
+		panic(err)
+	}
+	preimageCaveat := lsat.Caveat{Condition: lsat.PreimageKey, Value: preimage}
```
ellemouton, line 24: nit: lll

interceptor/client_interceptor.go
```diff
@@ -77,18 +78,18 @@ var (
...
+	store         lsat.Store
 	callTimeout   time.Duration
 	maxCost       btcutil.Amount
 	maxFee        btcutil.Amount
 	lock          sync.Mutex
 	allowInsecure bool
 }
 
-// NewInterceptor creates a new gRPC client interceptor that uses the provided
+// NewClientInterceptor creates a new gRPC client interceptor that uses the provided
```
ellemouton, line 89: nit: lll

ellemouton (commented):
Looking good!

The one thing im not sure of is the mocks. seems strange that the `auth` package would supply the mock for the `Mint` or that the `challenger` would supply the mock for the `InvoiceClient`. Also, im not sure how useful these mocks would be elsewhere - from my experience if I am using a mock, i mostly need to tweak it so that it works a certain way and that will be hard to do if importing the mock from another repo. Might be a case of "a little bit of copying is better than a little bit of dependency". Interested to hear other people's thoughts on this too though!

ellemouton:
!lightninglabs-deploy mute

PR #96, opened 2023-05-28
Add support for sqlite/postgres database backends

### LINKED ISSUE #95: db: add sqlite/postgres as a database backend option
Today we require `etcd`, which might be a bit more exotic to some devs. We should add support for sqlite+postgres via `sqlc` as we've done in some other projects. 

We only need to store the onion private key, and also the macaroon root keys. With taproot assets, we have an implementation of the macaroon storage, so we can use that: https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/macaroons.go

positiveblue:
This PR adds sqlite/postgres support for aperture. 

Currently the server uses a couple of interfaces `SecretStore` and `OnionStore` that are implemented in this PR. To keep consistency with other code bases aperture uses the same tools `sqlc`/`goang-migrate` and the same interfaces than  [Taproot Assets](https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/interfaces.go)

Fixes #95


aperturedb/sqlc/migrations/000001_secrets.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS secrets (
+    id INTEGER PRIMARY KEY,
+    hash BLOB NOT NULL,
+    secret BLOB NOT NULL,
+    created_at TIMESTAMP NOT NULL 
+);
+
+CREATE INDEX IF NOT EXISTS secrets_hash_idx ON secrets (hash);
+
```
Roasbeef, line 9: Extra new line?

aperturedb/sqlc/migrations/000002_onion.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS onion (
+    private_key BLOB NOT NULL,
```
Roasbeef, line 2: `UNIQUE`?

aperturedb/sqlc/migrations/000001_secrets.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS secrets (
+    id INTEGER PRIMARY KEY,
+    hash BLOB NOT NULL,
```
Roasbeef, line 3: `UNIQUE` constraint for both the hash and secret here?

aperturedb/interface.go
```diff
@@ -0,0 +1,145 @@
...
+
+	*sqlc.Queries
+}
+
+// BeginTx wraps the normal sql specific BeginTx method with the TxOptions
+// interface. This interface is then mapped to the concrete sql tx options
+// struct.
+func (s *BaseDB) BeginTx(ctx context.Context, opts TxOptions) (*sql.Tx, error) {
+	sqlOptions := sql.TxOptions{
+		ReadOnly: opts.ReadOnly(),
```
Roasbeef, line 142: Should tack this on just to be safe: https://github.com/lightninglabs/taproot-assets/blob/192c67ec33027da06343802ee6370bc1d79cb9ab/tapdb/interfaces.go#L239

Our transactions are simple enough that we shouldn't hit it, but why not add the extra level or protection.

aperturedb/postges.go
```diff
@@ -0,0 +1,131 @@
+package aperturedb
```
Roasbeef, line 1: Files name: postges -> postgres

aperturedb/postges.go
```diff
@@ -0,0 +1,131 @@
...
+		})
+
+		err = applyMigrations(
+			postgresFS, driver, "sqlc/migrations", cfg.DBName,
+		)
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
Roasbeef, line 103: Should set the connection limit like we do here: https://github.com/lightninglabs/taproot-assets/blob/main/tapdb/postgres.go#L75-L82

aperturedb/secrets.go
```diff
@@ -0,0 +1,171 @@
...
+	// InsertSecret inserts a new secret into the database.
+	InsertSecret(ctx context.Context, arg NewSecret) (int32, error)
+
+	// GetSecretByHash returns the secret that corresponds to the given
+	// hash.
+	GetSecretByHash(ctx context.Context, hash []byte) ([]byte, error)
+
+	// DeleteSecretByHash removes the secret that corresponds to the given
+	// hash.
+	DeleteSecretByHash(ctx context.Context, hash []byte) (int64, error)
```
Roasbeef, line 34: Do we actually need the returned primary key here? Or seems like a newer `sqlc` feature maybe?

aperturedb/secrets.go
```diff
@@ -0,0 +1,171 @@
...
+	db    BatchedSecretsDB
+	clock clock.Clock
+}
+
+// NewSecretsStore creates a new SecretsStore instance given a open
+// BatchedSecretsDB storage backend.
+func NewSecretsStore(db BatchedSecretsDB) *SecretsStore {
+	return &SecretsStore{
+		db:    db,
+		clock: clock.NewDefaultClock(),
```
Roasbeef, line 77: 👍

aperturedb/onion.go
```diff
@@ -0,0 +1,163 @@
...
+
+	var writeTxOpts OnionDBTxOptions
+	err := o.db.ExecTx(ctxt, &writeTxOpts, func(tx OnionDB) error {
+		// Only store the private key if it doesn't already exist.
+		_, err := tx.SelectOnionPrivateKey(ctxt)
+		if err != sql.ErrNoRows {
+			return fmt.Errorf("private key already exists")
+		}
+
+		fmt.Printf("inserting privatekey: %v\n", privateKey)
```
Roasbeef, line 92: Stray print statement?

aperturedb/onion.go
```diff
@@ -0,0 +1,163 @@
...
+// the OnionStore interface.
+func (o *OnionStore) StorePrivateKey(privateKey []byte) error {
+	ctxt, cancel := context.WithTimeout(
+		context.Background(), DefaultStoreTimeout,
+	)
+	defer cancel()
+
+	var writeTxOpts OnionDBTxOptions
+	err := o.db.ExecTx(ctxt, &writeTxOpts, func(tx OnionDB) error {
+		// Only store the private key if it doesn't already exist.
```
Roasbeef, line 86: Can be taken care of my making insert a no-op uspert. So would add that unique constraint, then catch the on conflict.

aperturedb/onion.go
```diff
@@ -0,0 +1,163 @@
...
+// ErrNoPrivateKey should be returned.
+func (o *OnionStore) PrivateKey() ([]byte, error) {
+	ctxt, cancel := context.WithTimeout(
+		context.Background(), DefaultStoreTimeout,
+	)
+	defer cancel()
+
+	var (
+		privateKey  []byte
+		writeTxOpts OnionDBTxOptions
```
Roasbeef, line 119: Can be a read tx.

config.go
```diff
@@ -98,6 +108,16 @@ type Config struct {
...
 	ServeStatic bool `long:"servestatic" description:"Flag to enable or disable static content serving."`
 
+	// DatabaseBackend is the database backend to be used by the server.
+	DatabaseBackend string `long:"databasebackend" description:"The database backend to use for storing all asset related data." choice:"sqlite" choice:"postgres"`
+
+	// Sqlite is the configuration section for the SQLite database backend.
+	Sqlite *aperturedb.SqliteConfig `group:"sqlite" namespace:"sqlite"`
+
+	// Postgres is the configuration section for the Postgres database backend.
+	Postgres *aperturedb.PostgresConfig `group:"postgres" namespace:"postgres"`
```
Roasbeef, line 118: 👍

Roasbeef (changes requested):
Nice work! Happy to see we could re-use so much existing code. Main comments are related to some schema modifications to simplify some of the biz logic, and make unique fields explicit.

aperturedb/interface.go
positiveblue: Better safe than sorry, I will make sure that we also add it to other projects (LND) :+1:

aperturedb/secrets.go
positiveblue: this does not return the primary key but the number of rows affected by this query. The idea is to be able to return an error if we are deleting a hash that is not currently in the db

aperturedb/onion.go
positiveblue: updated :+1:

aperturedb/postgres.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil, err
+	}
+
+	maxConns := defaultMaxConns
+	if cfg.MaxOpenConnections > 0 {
+		maxConns = int(cfg.MaxOpenConnections)
+	}
+
+	rawDB.SetMaxOpenConns(maxConns)
+	rawDB.SetMaxIdleConns(maxConns)
```
guggero, line 82: We need to actually set the number of maximum connections. Also, maybe it makes sense to take the other fixes applied here https://github.com/lightninglabs/taproot-assets/pull/336 and here https://github.com/lightninglabs/taproot-assets/pull/334

aperturedb/test_postgres.go
```diff
@@ -0,0 +1,13 @@
+//go:build test_db_postgres
+// +build test_db_postgres
```
guggero, line 2: Should we also run these unit tests in the CI?

guggero (commented):
Looks great, thanks a lot!
Found a couple of smaller things, but those should be relatively easy to address.

aperturedb/sqlc/queries/onion.sql
```diff
@@ -0,0 +1,16 @@
+-- name: UpsertOnion :exec
+INSERT INTO onion (
+    private_key, created_at
+) VALUES (
+    $1, $2
+) ON CONFLICT (
+    private_key
+) DO NOTHING;
```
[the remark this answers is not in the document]
positiveblue: yep, it works. As you say there in no difference because we do a select from the application and if there is already a row we do not allow inserting a new one. We can leave it as it is, I changed the test to reflect that we are able to upsert (when it matches what we already have in the db) without any problem.

aperturedb/test_postgres.go
positiveblue: totally, added to CI

aperturedb/postgres.go
positiveblue: `We need to actually set the number of maximum connections.` i do not know what this means. Aren't we doing that with `rawDB.SetMaxOpenConns(maxConns)`?

I also added the changes from tapdb, thanks for pointing to them :+1:

guggero: Sorry, my wording was a bit confusing. I meant that we have a config flag for Postgres that allows the user to specify the number of connections. But here we just set the constant, so the config flag is unused.

config.go
```diff
@@ -98,6 +108,16 @@ type Config struct {
 	// directory defined by StaticRoot.
 	ServeStatic bool `long:"servestatic" description:"Flag to enable or disable static content serving."`
 
+	// DatabaseBackend is the database backend to be used by the server.
+	DatabaseBackend string `long:"databasebackend" description:"The database backend to use for storing all asset related data." choice:"sqlite" choice:"postgres"`
```
Roasbeef, line 112: How about `db_backend` for the `long` here as it's shorter?

aperture.go
```diff
@@ -208,15 +211,75 @@ func (a *Aperture) Start(errChan chan error) error {
...
+		dbOnionTxer := aperturedb.NewTransactionExecutor(db,
+			func(tx *sql.Tx) aperturedb.OnionDB {
+				return db.WithTx(tx)
+			},
+		)
+		onionStore = aperturedb.NewOnionStore(dbOnionTxer)
+
+	default:
+		return fmt.Errorf("unknown database backend: %s",
+			a.cfg.DatabaseBackend)
```
Roasbeef, line 282: How about logging which DB backend is active?

aperture.go
```diff
@@ -208,15 +211,75 @@ func (a *Aperture) Start(errChan chan error) error {
...
+
+		dbOnionTxer := aperturedb.NewTransactionExecutor(db,
+			func(tx *sql.Tx) aperturedb.OnionDB {
+				return db.WithTx(tx)
+			},
+		)
+		onionStore = aperturedb.NewOnionStore(dbOnionTxer)
+
+	case "sqlite":
+		db, err := aperturedb.NewSqliteStore(a.cfg.Sqlite)
```
Roasbeef, line 259: Looks like if only `databasebackend: sqlite` is set, then the db file isn't set, so it ends up being `""`.

Roasbeef (changes requested):
Look preeetty good, ran things locally and found an issue with the `sqlite` file config if one isn't explicitly specified. I think we want to allow people to just say `db_backend: sqlite` and not have to explicitly set where the file should go (actually use the default).

config.go
positiveblue: none of the other config params separates its words. I like `db_backend` more but I think we should use `dbbackend` to be consistent

aperture.go
positiveblue: fixed

guggero (approved):
tACK, LGTM :tada: 

Tested in the Pool server integration test.

Roasbeef (approved):
LGTM 🛺

Roasbeef:
Needs a rebase!

PR #98, opened 2023-06-06
[feature] Use LNC to connect to the LND node

positiveblue:
Allow aperture to connect to LND using LNC


aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS sessions (
+    id INTEGER PRIMARY KEY,
+
+    passphrase_entropy BLOB NOT NULL UNIQUE,
+    remote_static_pub_key BLOB UNIQUE,
```
Roasbeef, line 5: This can be `NULL` before or after the handshake is complete?

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
```diff
@@ -0,0 +1,13 @@
...
+    local_static_priv_key BLOB NOT NULL,
+    mailbox_addr TEXT NOT NULL,
+
+    created_at TIMESTAMP NOT NULL,
+    expiry TIMESTAMP WITHOUT TIME ZONE
```
Roasbeef, line 10: Any reason why one needs the timezone info while the other doesn't?

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS sessions (
+    id INTEGER PRIMARY KEY,
+
+    passphrase_entropy BLOB NOT NULL UNIQUE,
+    remote_static_pub_key BLOB UNIQUE,
+    local_static_priv_key BLOB NOT NULL,
+    mailbox_addr TEXT NOT NULL,
+
```
Roasbeef, line 8: Maybe also a string label?

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
```diff
@@ -0,0 +1,13 @@
...
+    created_at TIMESTAMP NOT NULL,
+    expiry TIMESTAMP WITHOUT TIME ZONE
+);
+
+CREATE INDEX IF NOT EXISTS sessions_passphrase_entropy_idx ON sessions(passphrase_entropy);
```
Roasbeef, line 13: Index created here doesn't match the one made above.

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS sessions (
+    id INTEGER PRIMARY KEY,
+
+    passphrase_entropy BLOB NOT NULL UNIQUE,
```
Roasbeef, line 4: Is this the raw bytes, or the string pass phrase?

aperturedb/sqlc/queries/sessions.sql
```diff
@@ -0,0 +1,17 @@
...
+
+-- name: GetSessionByPassphrase :one
+SELECT *
+FROM sessions
+WHERE passphrase_entropy = $1;
+
+-- name: SetRemotePubKey :exec
+UPDATE sessions
+SET remote_static_pub_key=$1
+WHERE id=$2;
```
Roasbeef, line 16: Can alternatively be done via a label, or a hash of the entropy.

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
Roasbeef: Looks like the next commit assumes one exists.

lnc/lnc.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/mwitkow/grpc-proxy/proxy"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+// HeaderMacaroon is the HTTP header field name that is used to send
+// the macaroon.
+const HeaderMacaroon = "Macaroon"
+
+type conn struct {
```
Roasbeef, line 28: Should have `godoc` comments once the PR is out of draft.

lnc/lnc.go
```diff
@@ -0,0 +1,169 @@
...
+	cancel     func()
+}
+
+func (c *conn) Close() error {
+	err := c.grpcClient.Close()
+	c.cancel()
+	return err
+}
+
+type LNC struct {
```
Roasbeef, line 41: `lnc.NodeConnection`? 

`lnc.NodeConn`

`lnc.Node`

lnc/lnc.go
```diff
@@ -0,0 +1,169 @@
...
+
+	passphrase []byte
+	conn       *conn
+	mu         sync.Mutex
+}
+
+func NewLNC(store Store) *LNC {
+	return &LNC{
+		store: store,
+		mu:    sync.Mutex{},
```
Roasbeef, line 52: No need to init the mutex explicitly.

lnc/lnc.go
```diff
@@ -0,0 +1,169 @@
...
+
+	l.conn = conn
+	l.passphrase = passphrase
+
+	return nil
+}
+
+func (l *LNC) CloseConn() error {
+	l.mu.Lock()
+	defer l.mu.Unlock()
```
Roasbeef, line 98: One thing we'll want to test is that if the connection dies then we either retry, or we shut down.

lnc/lnc.go
```diff
@@ -0,0 +1,169 @@
...
+	)
+	defer cancel()
+
+	remoteKey := key.SerializeCompressed()
+
+	return l.store.SetRemotePubKey(ctxt, l.passphrase, remoteKey)
+}
+
+func (l *LNC) onAuthData(data []byte) error {
+	return nil
```
Roasbeef, line 138: IIRC, this is the call back that'll be used to get the macaroon in the latter portion of the first handshake? 

cc @ellemouton @guggero

lnc/session.go
```diff
@@ -0,0 +1,28 @@
+package lnc
+
+import (
+	"time"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+)
+
+// Session is a session
+type Session struct {
```
Roasbeef, line 10: As is, the definition and first usage is out of order in the commit structure.

challenger.go
```diff
@@ -66,23 +65,16 @@ const (
 	invoiceMacaroonName = "invoice.macaroon"
 )
 
-// NewLndChallenger creates a new challenger that uses the given connection
-// details to connect to an lnd backend to create payment challenges.
-func NewLndChallenger(cfg *AuthConfig, genInvoiceReq InvoiceRequestGenerator,
+// NewLndChallenger creates a new challenger that uses the given connection to
+// an lnd backend to create payment challenges.
+func NewLndChallenger(client InvoiceClient,
```
Roasbeef, line 70: 👍

lnc/lnc.go
```diff
@@ -48,17 +53,33 @@ func NewLNC(store Store) *LNC {
...
+			"words, got %d", mailbox.NumPassphraseWords,
+			len(words))
+	}
+
+	var mnemonicWords [mailbox.NumPassphraseWords]string
+	copy(mnemonicWords[:], words)
+	entropy := mailbox.PassphraseMnemonicToEntropy(mnemonicWords)
+
+	ctxb := context.Background()
+	session, err := l.store.GetSessionByPassphrase(ctxb, entropy[:])
```
Roasbeef, line 76: I think we want this to actually be an upsert? So we want the session struct, and it should return or insert depending on if it existed already or not.

Roasbeef (changes requested):

lnc/lnc.go
ellemouton: yeah

sample-conf.yaml
```diff
@@ -38,6 +38,18 @@ authenticator:
   # The chain network the lnd is active on.
   network: "simnet"
 
+# Settings for connecting to the LND node using LNC.
+lncauthenticator:
```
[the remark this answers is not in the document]
positiveblue: we usually do not use underscores in the config. I renamed it to `lnc`

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
positiveblue: only before the handshake is completed, when we create the session for first time (we do it before the connection)

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
positiveblue: not really, in postgres/sqlite `TIMESTAMP` is the same as `TIMESTAMP WITHOUT TIME ZONE`. I will remove it to be consistent with the shorter version `TIMESTAMP`

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
positiveblue: I wasn't sure about adding one but ended up adding it. 

A session is then identified by its label or its passphrase. Which is kind of weird because a passphrase is what defines the session. I added the field so it is more human friendly though.

aperturedb/sqlc/migrations/000003_lnc-sessions.up.sql
positiveblue: added comments to all fields

lnc/lnc.go
positiveblue: NodeConn it is :+1:

lnc/store.go
```diff
@@ -0,0 +1,31 @@
...
+	// entropy.
+	GetSession(ctx context.Context,
+		passphraseEntropy []byte) (*Session, error)
+
+	// SetRemotePubKey sets the remote public key for a session.
+	SetRemotePubKey(ctx context.Context, label string,
+		remotePubKey []byte) error
+
+	// SetExpiry sets the expiry time for a session.
+	SetExpiry(ctx context.Context, label string, expiry time.Time) error
```
ellemouton: still getting through the diff but what happens when the session expires? does aperture shutdown?

lnc/log.go
```diff
@@ -0,0 +1,26 @@
+package lnc
+
+import (
+	"github.com/btcsuite/btclog"
+	"github.com/lightningnetwork/lnd/build"
+)
+
+// Subsystem defines the sub system name of this package.
+const Subsystem = "LNCS"
```
ellemouton, line 9: what does the `S` stand for out of curiosity? 🤓

lnc/lnc.go
```diff
@@ -0,0 +1,331 @@
...
+	macBytes, err := mac.MarshalBinary()
+	if err != nil {
+		log.Errorf("unable to marshal macaroon for session(%v): %w",
+			l.session.Label, err)
+
+		return err
+	}
+
+	l.macStr = hex.EncodeToString(macBytes)
+
```
ellemouton, line 214: should we verify that the macaroon we get actually does have the permissions that aperture will need to operate?

aperturedb/sqlc/migrations/000003_lnc_sessions.up.sql
```diff
@@ -0,0 +1,29 @@
+-- lnc_sessions is table used to store data about LNC sesssions.
+CREATE TABLE IF NOT EXISTS lnc_sessions (
+    id INTEGER PRIMARY KEY,
+
+    -- The entropy bytes to be used for mask the local ephemeral key during the 
+    -- first step of the Noise XX handshake.
+    passphrase_entropy BLOB NOT NULL UNIQUE,
+
```
ellemouton, line 11: do we need to store this? we get it from the config file right? and then if the connection works then we dont need this again

ellemouton: ah ok yeah you are using it as an identifier. 

so I guess then my question is, why would we need more than 1 session?

pricesrpc/prices.swagger.json
```diff
@@ -73,14 +73,11 @@
     "protobufAny": {
       "type": "object",
       "properties": {
-        "type_url": {
+        "@type": {
```
ellemouton, line 76: separate commit for the non db changes maybe?

aperture.go
```diff
@@ -163,6 +164,7 @@ type Aperture struct {
 
 	etcdClient    *clientv3.Client
 	db            *sql.DB
+	lncNodeConn   *lnc.NodeConn
```
ellemouton, line 167: maybe makes sense to put inside the challenger instead?

Perhaps we make the challenger an interface and then let the LNC conn be one implementation of the Challenger? checkout what I did here for adding the LNURL challenger: https://github.com/lightninglabs/aperture/pull/77/commits/0c780a819b2b9d088c91f1989ec55babbc70f5da

ellemouton (changes requested):
Something needs to block calls until the connection is ready. Also maybe some info logs to say where the connection is at. 

Ran into this panic while testing. 
```
2023/06/30 11:36:31 http: panic serving 127.0.0.1:65481: runtime error: invalid memory address or nil pointer dereference                                                                                                                     [6/477]
goroutine 84 [running]:
net/http.(*conn).serve.func1()
        /usr/local/go/src/net/http/server.go:1850 +0xb0
panic({0x1056daf60, 0x106493e90})
        /usr/local/go/src/runtime/panic.go:890 +0x258
github.com/lightninglabs/aperture.(*LndChallenger).NewChallenge(0x0, 0x14000bec000?)
        /Users/elle/ll/aperture/challenger.go:269 +0x28
github.com/lightninglabs/aperture/mint.(*Mint).MintLSAT(0x140007929c0, {0x105927f08, 0x140001a2000}, {0x140003504e0, 0x1, 0x1})
        /Users/elle/ll/aperture/mint/mint.go:110 +0x68
github.com/lightninglabs/aperture/auth.(*LsatAuthenticator).FreshChallengeHeader(0x140008bc000, 0x1400029a700, {0x14000040700, 0x8}, 0xa)
        /Users/elle/ll/aperture/auth/authenticator.go:86 +0xb8
github.com/lightninglabs/aperture/proxy.(*Proxy).handlePaymentRequired(0x1400018e050, {0x105926fd0, 0x1400037e460}, 0x1400029a700, {0x14000040700, 0x8}, 0x1018?)
        /Users/elle/ll/aperture/proxy/proxy.go:403 +0x70
github.com/lightninglabs/aperture/proxy.(*Proxy).ServeHTTP(0x1400018e050, {0x105926fd0, 0x1400037e460}, 0x1400029a700)
        /Users/elle/ll/aperture/proxy/proxy.go:177 +0x2f8
net/http.HandlerFunc.ServeHTTP(0x14000080948?, {0x105926fd0?, 0x1400037e460?}, 0x40?)
        /usr/local/go/src/net/http/server.go:2109 +0x38
golang.org/x/net/http2/h2c.h2cHandler.ServeHTTP({{0x10591dca0?, 0x140003fe480?}, 0x1400018e140?}, {0x105926fd0, 0x1400037e460}, 0x1400029a700)
        /Users/elle/go/pkg/mod/golang.org/x/net@v0.10.0/http2/h2c/h2c.go:125 +0x41c
net/http.serverHandler.ServeHTTP({0x14000bacd20?}, {0x105926fd0, 0x1400037e460}, 0x1400029a700)
```

lnc/lnc.go
Roasbeef: Do we have methods exported from the `lnd` macaroons package or aperture to easily sanity check this?

aperturedb/sqlc/migrations/000003_lnc_sessions.up.sql
Roasbeef: I think so this can serve as the baiss for other projects/packages? So then they can have multiple sessions themselves, say a single `lncli` instance that can connect to all the other active sessions.

aperture.go
Roasbeef: +1 -- then also able to add other versions outside the package as well

aperture.go
```diff
@@ -293,7 +312,51 @@ func (a *Aperture) Start(errChan chan error) error {
...
+			return err
+		}
+
+		err = a.challenger.Start()
+		if err != nil {
+			return fmt.Errorf("unable to start lnd challenger: %w",
+				err)
+		}
+
+		a.lncNodeConn = nodeConn
```
Roasbeef, line 355: I think what we actually need is the ability to tear down the challenger? So we don't need to hold the connection, and can go with the interface approach @ellemouton mentioned above.

Roasbeef (changes requested):

config.go
```diff
@@ -49,6 +49,28 @@ type AuthConfig struct {
 	Disable bool `long:"disable" description:"Whether to disable LND auth."`
 }
 
+// LNC is the configuration section for connecting to the LND node using LNC.
+type LNC struct {
+	// Label is the label that will be used to identify the connection.
+	Label string `long:"label" description:"the label to be used for the connection"`
```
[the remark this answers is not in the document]
positiveblue: dropped

config.go
```diff
@@ -70,6 +92,29 @@ func (a *AuthConfig) validate() error {
...
+	if !l.Enable {
+		return nil
+	}
+
+	switch {
+	case l.Label == "":
+		return errors.New("lnc label required")
+
+	case l.Passphrase == "":
+		return errors.New("lnc passphrase required")
```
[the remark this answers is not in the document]
positiveblue: I added both to the db: words + entropy. Entropy is the main field use in the API but in the config file people need to use words (+ it can be useful in case people want to identify the entry in the db directly)

aperturedb/sqlc/migrations/000003_lnc_sessions.up.sql
positiveblue: We should not need more than one session, but we may need to update it because the old one expired, or we want to use a session with  more constraints for example. Right now there is no "session management" exposed to the user. We basically say "give me the passphrase that you want to use and i will use that to connect to the node"

In the example roasbeef is talking about it makes sense to use labels (so the user can connect to multiple nodes changing a short string in the params) but here I think the trade offs are different enough two  do not follow the same pattern in both cases.

lnc/log.go
positiveblue: Subsystem? I just did not know any other way to make it 4 letters, happy to use 3 or use another naming.

lnc/lnc.go
positiveblue: TODO added :+1:

lnc/lnc.go
positiveblue: If the connection dies, the invoice subscription returns an error, and we shutdown the server (in the same way we do if you are directly connecting to your LND node)

lnc/store.go
positiveblue: The next call will fail and the server will shutdown

aperture.go
positiveblue: Much better your way, thanks @ellemouton :tada:

lnc/lnc.go
```diff
@@ -0,0 +1,327 @@
...
+	return &conn{
+		client:     lnrpc.NewLightningClient(grpcClient),
+		grpcClient: grpcClient,
+		creds:      noiseConn,
+		cancel:     cancel,
+	}, nil
+}
+
+// extractMacaroon is a helper function that extracts a macaroon from raw bytes.
```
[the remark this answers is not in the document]
positiveblue:
```
Macaroon: hexData
```

Added :+1:

aperture.go
```diff
@@ -284,37 +289,32 @@ func (a *Aperture) Start(errChan chan error) error {
...
-		return &lnrpc.Invoice{
-			Memo:  "LSAT",
-			Value: price,
-		}, nil
-	}
-
 	if !a.cfg.Authenticator.Disable {
+		genInvoiceReq := func(price int64) (*lnrpc.Invoice, error) {
+			return &lnrpc.Invoice{
+				Memo:  "LSAT",
```
Roasbeef, line 312: Non-blocking, but in the future we can insert relevant endpoint/context information here so someone can build a UI/dashboard on top of lnd/aperture to do things like track the amount earned from a given endpoint, etc. 

To do this, we'd need to expose the memo as part of the challenger interface, as only much later in the pipeline do we have the full request context information.

challenger/lnd.go
```diff
@@ -263,7 +239,9 @@ func (l *LndChallenger) Stop() {
 // request (invoice) and the corresponding payment hash.
 //
 // NOTE: This is part of the mint.Challenger interface.
-func (l *LndChallenger) NewChallenge(price int64) (string, lntypes.Hash, error) {
+func (l *LndChallenger) NewChallenge(price int64) (string, lntypes.Hash,
```
Roasbeef, line 242: Re the above comment, we could pass in the additional context here as an optional argument. Then the invoice creation has a chance to tag any extra details.

config.go
```diff
@@ -37,16 +37,28 @@ type EtcdConfig struct {
...
 	MacDir string `long:"macdir" description:"Directory containing LND instance's macaroons"`
 
-	Network string `long:"network" description:"The network LND is connected to." choice:"regtest" choice:"simnet" choice:"testnet" choice:"mainnet"`
+	// The one-time-use passphrase used to set up the connection. This field
+	// identifies the connection that will be used.
+	Passphrase string `long:"passphrase" description:"the lnc passphrase"`
+
+	// MailboxAddress is the address of the mailbox that the client will
+	// use for the LNC connection.
+	MailboxAddress string `long:"mailboxaddress" description:"the host:port of the mailbox server to be used"`
```
Roasbeef, line 57: We should have a default mailbox addr we use here.

Roasbeef (approved):
LGTM 🔌

Fixed some issues I ran into in #103!

Issue #62, opened 2021-12-13
Upgrade to grpc v1.42.0 breaks client interceptor


carlaKC:
In lightninglabs/loop/pull/445, I bumped the following dependencies by mistake:

```
google.golang.org/grpc v1.38.0 -> google.golang.org/grpc v1.42.0
google.golang.org/protobuf v1.26.0 -> google.golang.org/protobuf v1.27.1
```

This caused all paywalled aperture apis to fail with `rpc error: code = Unknown desc = cannot initiate swap: unexpected HTTP status code received from server: 402 (Payment Required)`, despite the interceptor being active (confirmed with ad-hoc logging).

This is not an issue at present, since we have no pressing need to upgrade. 
Should we decide to upgrade lnd/loop/pool/lit, we will need to remember to upgrade aperture as well.

PR #102, opened 2023-06-15
proxy: always return 200 status code to gRPC client

guggero:
Fixes https://github.com/lightninglabs/aperture/issues/62.

It turns out that sending a non-200 HTTP status code was against the
gRPC spec and the older versions of the `grpc` library just didn't
validate that. The validation was added in v1.40.0, which is the version
that we couldn't update to before.
With this fix the error is still parsed correctly on the client side.
But this requires a small change to the L402 spec because the status
code is no longer 402.

See https://github.com/grpc/grpc-go/pull/4474 and https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses.

I'll also open a PR in the L402 spec repo and will test this in one of our server integration test setups.


proxy/proxy.go
```diff
@@ -437,7 +437,12 @@ func sendDirectResponse(w http.ResponseWriter, r *http.Request,
 		w.Header().Set(hdrGrpcStatus, strconv.Itoa(int(codes.Internal)))
 		w.Header().Set(hdrGrpcMessage, errInfo)
 
-		w.WriteHeader(statusCode)
+		// As per the gRPC spec, we need to send a 200 OK status code
```
Roasbeef, line 440: Ahh this makes a ton of sense, the spec also has this fragment shortly after: 
> Implementations should expect broken deployments to send non-200 HTTP status codes in responses as well as a variety of non-GRPC content-types and to omit Status & Status-Message. Implementations must synthesize a Status & Status-Message to propagate to the application layer when this occurs.

proxy/proxy.go
```diff
@@ -437,7 +437,12 @@ func sendDirectResponse(w http.ResponseWriter, r *http.Request,
...
+		// even if the request failed. The Grpc-Status and Grpc-Message
+		// header fields are enough to inform any gRPC compliant client
+		// about the error. See:
+		// https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses
+		w.WriteHeader(http.StatusOK)
```
Roasbeef, line 445: Will this be backwards compatible with the prior users of this package? I wonder if we should consider making this into a sub-module?

proxy/proxy.go
Roasbeef, line 440: Based on this: https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md

Should the above instead use an `UNKNOWN` status/error code?

Roasbeef (commented):
Nice investigation! 

One question here about backwards compat with other versions (sending 200 now each time vs the error code), and also about the HTTP error code to gRPC status mapping.

proxy/proxy.go
guggero: Yes, this should be backward compatible, since a gRPC client only looks at the `Grpc-Status` and `Grpc-Message` anyway. AFAIK this was always true, but newer clients just started validating the HTTP status code as well.
But this is what I aim to confirm with an itest and an older client.

proxy/proxy.go
guggero: Yes, probably. Or we could even send our own `Grpc-Status` code (for example 402). But because all clients out there explicitly check for the `codes.Internal` code at the moment, we would need to upgrade all clients first. Or parse the `User-Agent` field which contains the version of the `grpc` library. But that just seemed too brittle to me.

Roasbeef: Gotcha, let's stick with this for now then. When we refine the bLIP we can make sure to note these aspects for backwards compat.

Roasbeef (approved):
LGTM 🥠

I think the final thing we want to do here is upgrade some of our other deps that use Aperture in an itest to we have some regression protection/detection in those repos.

ellemouton (approved):
Wohoooo!!! 💥

guggero:
I verified that this works with both the Loop and Pool server backend integration tests :tada:

PR #105, opened 2023-07-05
config: set default mailbox address

positiveblue:
Make the LNC authenticator use the default mailbox address whenever the config does not specify a custom one.


Roasbeef (approved):
LGTM 🎉

Issue #114, opened 2023-10-23
POST request with body not working


sugarjig:
I have a very simple web service running locally. It has a POST endpoint at `/hello` that accepts a JSON body of the form

```json
{
  "name": "John"
}
```

and it responds with a body like

```json
{
  "message": "Hello John"
}
```

When I try using Aperture to proxy the request, I get an error. Here is the curl command I'm using

```sh
% curl -X POST -H "Content-Type: application/json" -d '{"name": "John"}' -k --http1.1 https://localhost:8081/hello
```

and the error I get is

```
curl: (18) transfer closed with 16 bytes remaining to read
```

However, if I don't send a body to the endpoint, I get a 402 as expected:

```sh
% curl -X POST -k --http1.1 https://localhost:8081/hello
payment required
```

FWIW, I am able to dump the request details right before the call to `sendDirectResponse` in the `handlePaymentRequired` method, and I can see the body I'm expecting.

```
POST /hello HTTP/1.1
Host: localhost:8081
Accept: */*
Access-Control-Allow-Headers: Authorization, Grpc-Metadata-macaroon, WWW-Authenticate
Access-Control-Allow-Methods: GET, POST, OPTIONS
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: WWW-Authenticate
Content-Length: 16
Content-Type: application/json
User-Agent: curl/8.1.2
Www-Authenticate: LSAT macaroon="...", invoice="..."

{"name": "John"}
```

Am I missing something in my request?

sugarjig:
Forgot to mention that when I try the same request in Postman, I get `Error: aborted`. Trying it in the JetBrains HTTP client, I get `org.apache.http.ConnectionClosedException: Premature end of Content-Length delimited message body (expected: 19; received: 17)`

guggero:
Hey. Thanks for the report. Looks like this was caused by the request headers being blindly copied when sending the L402 challenge header. Created a quick PR here: https://github.com/lightninglabs/aperture/issues/114. Will need to make sure nothing else breaks with that, but it should fix your issue.

sugarjig:
Thanks for the quick response!

PR #115, opened 2023-10-24
proxy+auth: fix post request

guggero:
Fixes https://github.com/lightninglabs/aperture/issues/114.

### DIFF
--- a/auth/authenticator.go
+++ b/auth/authenticator.go
@@ -83,8 +83,8 @@ const (
 // complete.
 //
 // NOTE: This is part of the Authenticator interface.
-func (l *L402Authenticator) FreshChallengeHeader(r *http.Request,
-	serviceName string, servicePrice int64) (http.Header, error) {
+func (l *L402Authenticator) FreshChallengeHeader(serviceName string,
+	servicePrice int64) (http.Header, error) {
 
 	service := l402.Service{
 		Name:  serviceName,
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -101,15 +101,19 @@ func TestProxyHTTP(t *testing.T) {
 	for _, tc := range testCases {
 		tc := tc
 
-		t.Run(tc.name, func(t *testing.T) {
-			runHTTPTest(t, tc)
+		t.Run(tc.name+" GET", func(t *testing.T) {
+			runHTTPTest(t, tc, "GET")
+		})
+
+		t.Run(tc.name+" POST", func(t *testing.T) {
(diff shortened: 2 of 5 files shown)


auth/authenticator.go
```diff
@@ -97,7 +97,7 @@ func (l *LsatAuthenticator) FreshChallengeHeader(r *http.Request,
 
 	str := fmt.Sprintf("LSAT macaroon=\"%s\", invoice=\"%s\"",
 		base64.StdEncoding.EncodeToString(macBytes), paymentRequest)
-	header := r.Header
```
Roasbeef, line 100: Why start from the blank header vs inheriting it from the request?

guggero: It didn't really make sense to start with the request header fields in the first place. Since the only (common) fields that are both valid in the request and response are the `Content-Type` and `Content-Length`. And the second one was what caused the issue.
So this was just wrong and we haven't noticed before.

Roasbeef: Gotcha, can we add a `POST` case to this test case so we're protected against regressions: https://github.com/lightninglabs/aperture/blob/master/proxy/proxy_test.go#L89

Roasbeef (changes requested):

guggero:
PR could be simplified after https://github.com/lightninglabs/aperture/pull/135, now it's just removing the unused parameter and adding a test case.

proxy/proxy_test.go
```diff
@@ -148,11 +152,29 @@ func runHTTPTest(t *testing.T, tc *testCase) {
...
 
 	require.Equal(t, "402 Payment Required", resp.Status)
 
+	bodyContent, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+	require.Equal(t, "payment required\n", string(bodyContent))
+
+	require.Equal(
+		t, fmt.Sprintf("%d", len(bodyContent)),
+		resp.Header.Get("Content-Length"),
```
starius, line 175: Field `resp.ContentLength` can be used, to avoid using `fmt.Sprintf("%d",`

Go documentation says:

>	// When
>	// Header values are duplicated by other fields in this struct (e.g.,
>	// ContentLength, TransferEncoding, Trailer), the field values are
>	// authoritative.

https://pkg.go.dev/net/http#Response

There are two other instances below in the code.

starius (approved):
LGTM 🎉

Thank you for refactoring it!

proxy/proxy_test.go
guggero: Makes sense, thanks! Fixed.

PR #120, opened 2023-11-16
config: fix profile port arg parsing

Roasbeef:
In this commit, we fix a subtle bug in the parsing of the yaml config. With the way the library works, the attribute name needs to match the config attribute name. Otherwise, parsing just doesn't work.

### DIFF
--- a/config.go
+++ b/config.go
@@ -200,8 +200,8 @@ type Config struct {
 	// BaseDir is a custom directory to store all aperture flies.
 	BaseDir string `long:"basedir" description:"Directory to place all of aperture's files in."`
 
-	// ProfilePort is the port on which the pprof profile will be served.
-	ProfilePort uint16 `long:"profile" description:"Enable HTTP profiling on given port -- NOTE port must be between 1024 and 65535"`
+	// Profile is the port on which the pprof profile will be served.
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -200,8 +200,8 @@ type Config struct {
 	// BaseDir is a custom directory to store all aperture flies.
 	BaseDir string `long:"basedir" description:"Directory to place all of aperture's files in."`
 
-	// ProfilePort is the port on which the pprof profile will be served.
-	ProfilePort uint16 `long:"profile" description:"Enable HTTP profiling on given port -- NOTE port must be between 1024 and 65535"`
+	// Profile is the port on which the pprof profile will be served.
+	Profile uint16 `long:"profile" description:"Enable HTTP profiling on given port -- NOTE port must be between 1024 and 65535"`
```
ellemouton, line 204: ah! interesting - TIL

ellemouton (approved):
great catch! 🔍

config.go
Roasbeef: Yeah a weird quirk with the YAML parser, I think we might've ran into this w/ the sqlite stuff too, but alas the wisdom was forgotten....

PR #127, opened 2024-02-10
loc: Add a timeout when creating the grpc client with the mailbox

### LINKED ISSUE #124: No response when dialing to Mailbox via gPRC
There is no error or cancellation at calling `grpc.DialContext` below, it is just put on hold. 
The mailbox is `mailbox.terminal.lightning.today:443`
https://github.com/lightninglabs/aperture/blob/master/lnc/lnc.go#L284-L286

By the way, this issue may be due to be re-use passphrase from multiple containers under development, because it worked generating new phrase per container. However, I don't know because I haven't received a response and seen any documentation prohibiting passphrase reuse.

positiveblue:
Do not hang forever when creating the grpc client with the mailbox.

Also, make sure we use the `DefaultStore` timeouts every time we call store functions in the package.

Both timeouts are currently set to 10s.

Fix #124


lnc/lnc.go
```diff
@@ -281,8 +281,12 @@ func (n *NodeConn) newConn(session *Session, opts ...grpc.DialOption) (*conn,
 	}
 	dialOpts = append(dialOpts, opts...)
 
+	ctxt, cancelT := context.WithTimeout(
+		context.Background(), DefaultConnectionTimetout,
```
guggero, line 289: Since we already have `ctxc` from above, shouldn't we use that as the parent context?

lnc/lnc.go
```diff
@@ -281,8 +281,12 @@ func (n *NodeConn) newConn(session *Session, opts ...grpc.DialOption) (*conn,
...
+	)
+	defer cancelT()
```
guggero, line 291: Are you sure we can cancel this context here? Is the context only used for the initial dialing and not for the whole duration of the connection?

guggero (commented):
Thanks for the fix! Code looks good, just a question about context lifetime.

lnc/lnc.go
positiveblue: Good question, I had to check it before opening the PR because I wasn't sure neither.


Form the [grpc docs](https://pkg.go.dev/google.golang.org/grpc#DialContext):

> In the blocking case, ctx can be used to cancel or expire the pending connection. Once this function returns, the cancellation and expiration of ctx will be noop. Users should call ClientConn.Close to terminate all the pending operations after this function returns.

Deployed and tested on Friday  👍

guggero:
@positiveblue the code looks good, thanks. Maybe we should also bump the LNC dependency to the recently released version `v0.3.0` that fixes a couple of issues. That probably will also help increasing the connection stability and reliability.

positiveblue:
`lightning-node-connect` version bumped :+1:


I also had to bump the minim go version for compiling the project because LNC needs `go 1.21`

guggero (approved):
utACK, LGTM :tada:

PR #128, opened 2024-02-13
sqlc: bump version to `v1.25.0`

positiveblue:
Since v0.18.0 there they multiple bug fixes, support more postgres/sqlite features, etc...

Nothing impacting the current code.

NOTE: It also looks like sqlc is growing as a company and wants to support more languages, add feature for cloud clients, etc... Because of the open source nature of this project I do not think that brings any extra benefit. The only remarkable thing is that they are working on doing the code more modular, and extracting the code generation in multiple "plugins" They already extracted the go one. No changes are needed by now, but maybe at some point they delete the code gen form the main project and the config files need to be updated to use the plugin system [sqlc-dev/plugin-sdk-go].


guggero (approved):
Thanks, LGTM :tada:

bhandras (approved):

PR #139, opened 2024-06-21
Bump lnd compile time dependency to v0.17.0, update RPC compilation of test proto

guggero:
Replaces https://github.com/lightninglabs/aperture/pull/138.

PR #140, opened 2024-07-08
pricesrpc: fix build tag formatting

guggero:
Fixes https://github.com/lightninglabs/aperture/pull/139#issuecomment-2182902879, didn't realize that no change to `falafel` itself was needed...


starius (approved):
LGTM 🎉

I tested `make fmt` locally. It does not produce any diff.
The test failure in CI seem to be flaky.

Issue #143, opened 2024-07-27
Not following protocol specification for multiple macaroons


gofeuer:
The protocol specification [here](https://github.com/lightninglabs/L402/blob/fda63dd8f39b997023d0e02638f8dfae34f7a456/protocol-specification.md) says that the macaroon portion of the authentication token, left side of `:` may contain multiple comma separated Macaroons.

Trying to send multiple comma separated macaroons will fail on line 72:
https://github.com/lightninglabs/aperture/blob/a30875823faa476a513f89a200899d4959bf2a2f/l402/header.go#L70-L76

I think this might actually be a specification problem and not a bug here.
Because what would be the use case for multiple macaroons to be passed to a request? They are all paid for my one invoice?

guggero:
Thanks for the report!

I think there might be use cases, where you'll want to give out multiple macaroons with different base permissions, but all tied to the same payment hash (invoice). So I think we should update the code to allow for multiple macaroons.

gofeuer:
Oh right, then `minter.MintL402(...)` should return a slice of macaroons here:

https://github.com/lightninglabs/aperture/blob/a30875823faa476a513f89a200899d4959bf2a2f/auth/authenticator.go#L94-L96

That would then be encoded with comma separators if the slice has more than one macaroon here:

https://github.com/lightninglabs/aperture/blob/a30875823faa476a513f89a200899d4959bf2a2f/auth/authenticator.go#L110-L111

This will allow a use case where the minter provides a full permission macaroon followed by ever more attenuated versions of it with the properly set caveats. Allowing the client to respond with one of the received macaroons that the caveats grant enough access to the resource they require. 

I get the client receiving multiple macaroons. But isn't allowing for them to provide multiple macaroons to authenticate a request an unnecessary  complication?

guggero:
>  But isn't allowing for them to provide multiple macaroons to authenticate a request an unnecessary complication?

I don't think it's too much of a hassle. We'd just validate more than one macaroon. But they would all need to be valid at the same time.

I don't remember exactly why we wrote the spec that way (it's been a while), but since it's quite explicit I think we might have had a specific use case in mind. So I think a first step would be to just _validate_ multiple macaroons, changing this code as you mentioned initially: https://github.com/lightninglabs/aperture/blob/a30875823faa476a513f89a200899d4959bf2a2f/l402/header.go#L70-L76

[...]

PR #146, opened 2024-08-11
Accept comma separated macaroons

gofeuer:
#143 
This will allow Aperture to receive comma separated macaroons and not reject the request as an error.
If multiple macaroons are received, the first one is used and the others are ignored (for now)

### DIFF
--- a/auth/authenticator.go
+++ b/auth/authenticator.go
@@ -41,11 +41,14 @@ func (l *L402Authenticator) Accept(header *http.Header, serviceName string) bool
 	// Try reading the macaroon and preimage from the HTTP header. This can
 	// be in different header fields depending on the implementation and/or
 	// protocol.
-	mac, preimage, err := l402.FromHeader(header)
+	macs, preimage, err := l402.FromHeader(header)
 	if err != nil {
 		log.Debugf("Deny: %v", err)
 		return false
 	}
+
+	// TODO: Be able to accept multiple macaroons
+	mac := macs[0]
 
 	verificationParams := &mint.VerificationParams{
 		Macaroon:      mac,
--- a/l402/header.go
+++ b/l402/header.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"strings"
 
 	"github.com/lightningnetwork/lnd/lntypes"
 	"gopkg.in/macaroon.v2"
@@ -42,7 +43,7 @@ var (
 //
 // If only the macaroon is sent in header 2 or three then it is expected to have
 // a caveat with the preimage attached to it.
-func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, error) {
+func FromHeader(header *http.Header) (macaroon.Slice, lntypes.Preimage, error) {
 	var authHeader string
 
 	switch {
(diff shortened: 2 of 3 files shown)


l402/header.go
```diff
@@ -71,14 +72,22 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
 		macBase64, preimageHex := matches[2], matches[3]
 		macBytes, err := base64.StdEncoding.DecodeString(macBase64)
 		if err != nil {
-			return nil, lntypes.Preimage{}, fmt.Errorf("base64 "+
-				"decode of macaroon failed: %v", err)
+			// macBase64 might be multiple macaroons separated by commas,
```
guggero, line 75: nit: comments should always be full sentences, including punctuation.

I also think this needs some more explanation that we're using `macaroon.Slice` below, which can actually decode multiple macaroons that are in one continuous blob of data.

l402/header.go
```diff
@@ -71,14 +72,22 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+			// so we strip them and try again
+			macBase64 = strings.ReplaceAll(macBase64, ",", "")
+			macBytes, err = base64.StdEncoding.DecodeString(macBase64)
+			if err != nil {
+				return nil, lntypes.Preimage{},
```
guggero, line 80: nit: please break the lines before 80 characters, while configuring your IDE to show the tabulator character as 8 spaces (most of the code in this block is now too long).

l402/header.go
```diff
@@ -71,14 +72,22 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+					fmt.Errorf("base64 decode of macaroon failed: %v", err)
+			}
 		}
-		mac := &macaroon.Macaroon{}
+		// macBytes can contain multiple macaroons,
```
guggero, line 84: nit: add empty line before comment. Also missing full stop.

l402/header.go
```diff
@@ -71,14 +72,22 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+		// (* macaroon.Slice) knows how to decode them without separators
+		mac := make(macaroon.Slice, 0, 1)
 		err = mac.UnmarshalBinary(macBytes)
 		if err != nil {
-			return nil, lntypes.Preimage{}, fmt.Errorf("unable to "+
-				"unmarshal macaroon: %v", err)
+			return nil, lntypes.Preimage{},
+				fmt.Errorf("unable to unmarshal macaroon: %v", err)
```
guggero, line 90: Why this formatting change here and below?

l402/header.go
```diff
@@ -107,32 +116,39 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
 		return nil, lntypes.Preimage{}, fmt.Errorf("unable to "+
 			"unmarshal macaroon: %v", err)
 	}
-	preimageHex, ok := HasCaveat(mac, PreimageKey)
-	if !ok {
-		return nil, lntypes.Preimage{}, errors.New("preimage caveat " +
-			"not found")
+	var preimageHex string
+	var ok bool
+	for _, m := range mac {
```
guggero, line 130: I think if there are multiple macaroons, then _all_ of them should have the caveat. Not just one of them. Otherwise IMO the whole idea of having multiple macaroons is weird.

l402/header.go
```diff
@@ -107,32 +116,39 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
 	err = mac.UnmarshalBinary(macBytes)
 	if err != nil {
 		return nil, lntypes.Preimage{}, fmt.Errorf("unable to "+
 			"unmarshal macaroon: %v", err)
 	}
-	preimageHex, ok := HasCaveat(mac, PreimageKey)
-	if !ok {
-		return nil, lntypes.Preimage{}, errors.New("preimage caveat " +
-			"not found")
+	var preimageHex string
```
guggero, line 128: nit: multiple var statements can be combined into one:
```
var (
    ...
)
```

l402/header.go
```diff
@@ -107,32 +116,39 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+	var ok bool
+	for _, m := range mac {
+		if preimageHex, ok = HasCaveat(m, PreimageKey); ok {
+			break
+		}
+	}
+	if preimageHex == "" || !ok {
```
guggero, line 135: We probably also want to make sure all macaroons reference the same preimage.

auth/authenticator.go
```diff
@@ -41,12 +41,15 @@ func (l *L402Authenticator) Accept(header *http.Header, serviceName string) bool
...
 	// be in different header fields depending on the implementation and/or
 	// protocol.
-	mac, preimage, err := l402.FromHeader(header)
+	macs, preimage, err := l402.FromHeader(header)
 	if err != nil {
 		log.Debugf("Deny: %v", err)
 		return false
 	}
 
+	// TODO: Be able to accept multiple macaroons
```
guggero, line 50: nit: add full stop to end of sentence. Here and in `server_interceptor.go`.

l402/header.go
```diff
@@ -107,32 +116,39 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
 	// extract the preimage.
 	macBytes, err := hex.DecodeString(authHeader)
 	if err != nil {
-		return nil, lntypes.Preimage{}, fmt.Errorf("hex decode of "+
-			"macaroon failed: %v", err)
+		return nil, lntypes.Preimage{},
+			fmt.Errorf("hex decode of macaroon failed: %v", err)
 	}
-	mac := &macaroon.Macaroon{}
+	mac := make(macaroon.Slice, 0, 1)
```
guggero, line 122: I know there isn't a unit test now, but it would be really nice to add one so this could be tested.

guggero (commented):
Thanks for the feature!
I think the approach is good. Most of my comments are around code style. I know this isn't specified explicitly in this project, but we try to adhere to the same [formatting rules](https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md) in all our projects.

guggero:
!lightninglabs-deploy mute

PR #148, opened 2024-09-11
slog: update to use structured logs for hashmail server

ellemouton:
Update aperture's btclog and lnd deps so that it can take advantage of structure logging. 

Then start using the structured logs for the hashmail server. 

This will make querying logs much easier & faster. 

## Example

Main thing to note here is the uniformity of the "stream_id" attribute in the new logs.

Old: 

Screenshot 2024-09-11 at 17 16 56

New:

Screenshot 2024-09-11 at 17 16 15


hashmail_server.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/btcsuite/btclog"
```
guggero, line 7: nit: need to `make fmt`.

hashmail_server.go
```diff
@@ -568,16 +574,16 @@ func (h *hashMailServer) NewCipherBox(ctx context.Context,
...
-		init.Desc.StreamId, init.Auth)
+	ctxl := btclog.WithCtx(ctx, btclog.Hex("stream_id", init.Desc.StreamId))
 
-	if err := h.ValidateStreamAuth(ctx, init); err != nil {
-		log.Debugf("Stream creation validation failed (id=%x): %v",
-			init.Desc.StreamId, err)
+	log.DebugS(ctxl, "New HashMail stream init", "auth", init.Auth)
+
+	if err := h.ValidateStreamAuth(ctxl, init); err != nil {
+		log.DebugS(ctxl, "Stream creation validation failed", err)
```
guggero, line 582: I think this needs to be `"err", err`, otherwise this would show as "bad key", because `DebugS` doesn't have an `err` parameter. Same further below.

guggero (approved):
Very nice! And cool to see a demo of the context based trace functionality :100: 

LGTM pending the dependent PRs and the `err` value fix.

hashmail_server.go
ellemouton: a yes good catch! 

makes me wonder if we should switch this to a "WarnS"? but anyways, keeping it as debug for now 👍

sample-conf.yaml
```diff
@@ -215,3 +215,12 @@ hashmail:
...
   listenaddr: "localhost:9000"
+
+# Console and file logger settings.
+logging:
+  console:
+    disable: false
+    call-site: short
+  file:
+    disable: true
+    call-site: long
```
ellemouton, line 226: yaml parser is struggling with these 🙃

guggero:
I think we need to bump the Golang version in the linter docker file.

sample-conf.yaml
guggero: Could be reserved words. Try `'short' and 'long' instead?

guggero (approved):
Very nice, LGTM :tada:

ellemouton:
struggling with this linter 🙃  will try some iterations locally

sample-conf.yaml
ellemouton: it isnt just the call-site. it is anything from that unexported empedded struct. only works for the 'style' option when compiled in 'dev' mode. 

seems to be a yaml issue (cause things work for the lnd.conf file just fine)

ellemouton:
ooo ok im winning now! gimme a min

ellemouton:
YAY linter happy

starius (approved):
LGTM!
Added few comments.

sample-conf.yaml
ellemouton: fixed here: https://github.com/lightningnetwork/lnd/pull/9576

aperture.go
```diff
@@ -619,7 +619,10 @@ func setupLogging(cfg *Config, interceptor signal.Interceptor) error {
 	}
 
 	// Now initialize the logger and set the log level.
-	SetupLoggers(logWriter, interceptor)
+	sugLogMgr = build.NewSubLoggerManager(
+		build.NewDefaultLogHandlers(cfg.Logging, logWriter)...,
+	)
```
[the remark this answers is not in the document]
ellemouton: indeed :) removed!

ellemouton (commented):
Thanks yall 🙏 
Updated.

Got the yaml parsing working but need to get https://github.com/lightningnetwork/lnd/pull/9576 merged first and then point to that here before merge

ellemouton:
oh - I've also bumped our Yaml lib btw since the old one is not actively maintained

PR #149, opened 2024-09-18
mod+aperture: bump lnd compile time dependency to latest master

guggero:
Due to a change in the logging library of lnd, we need to pass in a new parameter to be compatible with the latest lnd master version.


bhandras (approved):

jharveyb (approved):
LGTM 👍

PR #160, opened 2025-01-13
challenger: paginage ListInvoices to avoid resource exhaustion

bhandras:


challenger/lnd.go
```diff
@@ -84,49 +89,61 @@ func (l *LndChallenger) Start() error {
...
+			}
+			l.invoiceStates[hash] = invoice.State
 		}
+		l.invoicesMtx.Unlock()
 
-		// Don't track the state of canceled or expired invoices.
-		if invoiceIrrelevant(invoice) {
-			continue
+		// If there are no more invoices, stop pagination.
+		if len(invoiceResp.Invoices) < invoiceQueryPageSize {
```
guggero, line 140: This seems to work with the current implementation of the invoice paginator that always returns exactly the maximum number of invoices. But on an index based query where we might have "holes" from deleted items, this would abort too early.
So with the risk of making one query too much, should we just check `len(invoiceResp.Invoices) == 0`?

guggero (commented):
Thanks for the quick fix! One question.

challenger/lnd.go
bhandras: Yeah good idea! Done.

challenger/lnd.go
```diff
@@ -84,49 +89,61 @@ func (l *LndChallenger) Start() error {
...
-	// updates for new invoices and/or newly settled invoices.
-	l.invoicesMtx.Lock()
-	for _, invoice := range invoiceResp.Invoices {
-		// Some invoices like AMP invoices may not have a payment hash
-		// populated.
-		if invoice.RHash == nil {
-			continue
+	indexOffset := uint64(0)
+	for {
+		invoiceResp, err := l.client.ListInvoices(
```
guggero, line 98: Could also add some debug logging here so we could observe the startup progress in the logs.

guggero (approved):
utACK, LGTM :tada:

PR #172, opened 2025-05-06
aperture: internally aggregate session statistics to reduce cardinality

djkazic:
This PR adds aggregation and classification of sessions into `{active, standby, inuse}` categories and makes direct metrics exportable.

It is an attempt to lower the cardinality of the prior used read count metric. This should greatly reduce the load on Prometheus for both storage and querying.

However, there is a caveat with the code as-is:
* Unbounded memory growth: while the prior approach essentially made it the prometheus server's problem, we now have to deal with this ourselves. Looking for feedback on an appropriate culling threshold. I wanted to first and foremost mimic what we were doing before so I did not implement culling yet.

### DIFF
--- a/hashmail_server.go
+++ b/hashmail_server.go
@@ -11,7 +11,6 @@ import (
 	"github.com/btcsuite/btclog/v2"
 	"github.com/lightninglabs/lightning-node-connect/hashmailrpc"
 	"github.com/lightningnetwork/lnd/tlv"
-	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/time/rate"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -35,6 +34,11 @@ const (
 	// DefaultBufSize is the default number of bytes that are read in a
 	// single operation.
 	DefaultBufSize = 4096
+
+	// streamTTL is the amount of time that a stream needs to be exist without
+	// reads for it to be considered for pruning. Otherwise, memory will grow
+	// unbounded.
+	streamTTL = 24 * time.Hour
 )
 
 // streamID is the identifier of a stream.
@@ -747,9 +751,7 @@ func (h *hashMailServer) RecvStream(desc *hashmailrpc.CipherBoxDesc,
 		streamID := newStreamID(desc.StreamId)
 		if streamID.isOdd() {
 			baseID := streamID.baseID()
-			mailboxReadCount.With(prometheus.Labels{
-				streamIDLabel: fmt.Sprintf("%x", baseID),
-			}).Inc()
+			streamActivityTracker.Record(fmt.Sprintf("%x", baseID))
 		}
 
 		err = reader.Send(&hashmailrpc.CipherBox{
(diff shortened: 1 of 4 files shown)


djkazic:
Pushing a lint fix.

prometheus.go
```diff
@@ -3,35 +3,44 @@ package aperture
...
 	})
 
-	// mailboxReadCount counts each time a mailbox pair is being used.
-	// A session consists of a bidirectional stream each using a mailbox
-	// with an ID that overlaps for the first 63 bytes and differ for the
-	// last bit. So in order to obtain accurate data about a specific
-	// mailbox session, the stream ID that will be recorded is the first
-	// 16 bytes of the session ID and we will only record the odd stream's
-	// reads so that we don't duplicate the data.
-	mailboxReadCount = prometheus.NewCounterVec(
```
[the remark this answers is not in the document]
djkazic: Yes, however if we purge these old values that does represent a departure from our prior calculation methods. Specifically:

Say we have a 10 minute TTL, and a stream goes quiet.
1. We won’t track a stream’s activity anymore unless it becomes active again and is explicitly re-`Record()`ed.
2. If a stream becomes active again after pruning, its counter will be treated as "new" (i.e., no memory of prior rate).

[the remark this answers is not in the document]
djkazic: Will check this tomorrow morning.

hashmail_server.go
```diff
@@ -766,6 +763,76 @@ func (h *hashMailServer) RecvStream(desc *hashmailrpc.CipherBoxDesc,
...
+	defer sa.Unlock()
+
+	now := time.Now()
+
+	for _, e := range sa.streams {
+		elapsed := now.Sub(e.lastUpdate).Seconds()
+		if elapsed <= 0 {
+			elapsed = 1 // Prevent divide-by-zero, treat as 1s interval
+		}
+		rate := float64(e.count) / elapsed
```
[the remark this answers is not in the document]
djkazic: This code simply reimplements the logic we use in our dashboard PromQL, only reimplemented in Go.

https://lightninglabs.grafana.net/d/degu2d46pjnr4f/lnc-mailbox-server?orgId=1&from=now-6h&to=now&timezone=browser&var-datasource=default&editPanel=2

hashmail_server.go
```diff
@@ -766,6 +763,76 @@ func (h *hashMailServer) RecvStream(desc *hashmailrpc.CipherBoxDesc,
...
+func (sa *streamActivity) ClassifyAndReset() (active, standby, inuse int) {
+	sa.Lock()
+	defer sa.Unlock()
+
+	now := time.Now()
+
+	for _, e := range sa.streams {
+		elapsed := now.Sub(e.lastUpdate).Seconds()
+		if elapsed <= 0 {
+			elapsed = 1 // Prevent divide-by-zero, treat as 1s interval
```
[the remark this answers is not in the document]
djkazic: Will do a pass on style once we have the other aspects nailed down for the PR.

djkazic:
Force pushed some changes around TTL and formatting / comments. PTAL @Roasbeef

Roasbeef (approved):
LGTM ❄️

prometheus.go
```diff
@@ -3,35 +3,44 @@ package aperture
...
-		prometheus.CounterOpts{
-			Namespace: "hashmail",
-			Name:      "mailbox_read_count",
-		}, []string{streamIDLabel},
-	)
+	// activeSessions tracks the active session count for mailbox
```
guggero, line 19: nit: please finish all sentences with punctuation.

prometheus.go
```diff
@@ -55,7 +64,22 @@ func StartPrometheusExporter(cfg *PrometheusConfig) error {
...
 	prometheus.MustRegister(mailboxCount)
-	prometheus.MustRegister(mailboxReadCount)
+	prometheus.MustRegister(activeSessions)
+	prometheus.MustRegister(standbySessions)
+	prometheus.MustRegister(inUseSessions)
+
+	// Periodically update session classification metrics from internal tracker
+	go func() {
+		ticker := time.NewTicker(10 * time.Second)
+		defer ticker.Stop()
```
guggero, line 74: This Goroutine will never really stop, so the `defer` will never be called. Perhaps we should pass in the `interceptor` that we create in `run()` in `aperture.go`. Then at least we can listen on `interceptor.ShutdownChannel()`.

guggero (commented):
Thanks for looking into this.
Were you able to test this with the [LNC test page](https://github.com/lightninglabs/lightning-node-connect/tree/master/example)?
Would perhaps make sense to make sure the numbers will be what we expect. Maybe @ViktorTigerstrom has time to do a test run?

prometheus.go
djkazic: Will fix this.

prometheus.go
djkazic: Great catch. I'll refactor this after standup.

djkazic: Pushed a commit that should address this. Also fixed up the comment punctuation across the board.

Will rebase and squash before merge if everything looks ok.

guggero (approved):
utACK, LGTM :tada:

PR #175, opened 2025-05-12
multi: add new strict-verify CLI option to control invoice usage

Roasbeef:
In this commit, we add a new CLI argument that allows a user to control if we use strict verification or not. Strict verification relies on checking the actual invoice state against lnd, and requires more state for the Aperture server.

When strict verification isn't on, we rely only on the preimage payment hash relationship. Namely that the only way a user can obtain the preimage is to pay the invoice, and as we check the HMAC on the macaroon, we know that we created it with an invoice obtained from lnd.

### DIFF
--- a/config.go
+++ b/config.go
@@ -20,6 +20,7 @@ var (
 	defaultLogLevel         = "info"
 	defaultLogFilename      = "aperture.log"
 	defaultInvoiceBatchSize = 100000
+	defaultStrictVerify     = false
 
 	defaultSqliteDatabaseFileName = "aperture.db"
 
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -224,6 +225,11 @@ type Config struct {
 	// request.
 	InvoiceBatchSize int `long:"invoicebatchsize" description:"The number of invoices to fetch in a single request."`
 
+	// StrictVerify is a flag that indicates whether we should verify the
+	// invoice status strictly or not. If set to true, then this requires
+	// all invoices to be read from disk at start up.
+	StrictVerify bool `long:"strictverify" description:"Whether to verify the invoice status strictly or not."`
```
guggero, line 231: Should add this with the same comment to the example YAML config as well for documentation.

guggero (approved):
Nice, LGTM :tada:

config.go
Roasbeef: Sure, I added some other options that weren't updated here in a new commit.

Roasbeef:
Rebased.

bhandras (approved):
LGTM, thank you! 🎉

PR #180, opened 2025-05-27
challanger: fix shutdown panic

hieblmi:
Seeing the following because `l.invoiceCancel()` isn't initialized when config `strictverify` isn't set.
```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x103fcf340]

goroutine 1 [running]:
github.com/lightninglabs/aperture/challenger.(*LndChallenger).Stop(0x140005fbc00)
        /Users/bob/projects/go/src/github.com/aperture/challenger/lnd.go:273 +0x20
```

### DIFF
--- a/challenger/lnd.go
+++ b/challenger/lnd.go
@@ -270,7 +270,9 @@ func (l *LndChallenger) readInvoiceStream(
 
 // Stop shuts down the challenger.
 func (l *LndChallenger) Stop() {
-	l.invoicesCancel()
+	if l.invoicesCancel != nil {
+		l.invoicesCancel()
+	}
 	close(l.quit)
 	l.wg.Wait()
 }


challenger/lnd.go
```diff
@@ -270,7 +270,9 @@ func (l *LndChallenger) readInvoiceStream(
 
 // Stop shuts down the challenger.
 func (l *LndChallenger) Stop() {
-	l.invoicesCancel()
+	if l.strictVerify {
```
guggero, line 273: Should we instead check if `l.invoicesCancel` is not nil? In case it wasn't created for another reason, independent of the `strictVerify` flag.

guggero (approved):
Thanks for the fix!

PR #195, opened 2025-08-22
proxy: a configurable rate-limiter per endpoint and L402

hieblmi:
Implements a per-endpoint token-bucket rate limiter for aperture to protect against DoS attacks from authenticated or unauthenticated users (addresses #200).

  Key features:
  - Token-bucket algorithm using golang.org/x/time/rate
  - Per-L402 token ID limiting with IP fallback for unauthenticated requests
  - LRU cache (10k entries default) to prevent memory exhaustion
  - All matching rules enforced (strictest wins)
  - HTTP 429 + Retry-After header for REST, ResourceExhausted for gRPC
  - Prometheus metrics: aperture_ratelimit_{allowed,denied}_total, cache_size, evictions_total

  Configuration:
```
  services:
    - name: "my-service"
      ratelimits:
        - pathregexp: '^/api/v1/.*$'
          requests: 100
          per: 1m
          burst: 150
```

### DIFF
--- a/proxy/ratelimiter.go
+++ b/proxy/ratelimiter.go
@@ -43,7 +43,8 @@ func (e *limiterEntry) Size() (uint64, error) {
 
 // RateLimiter manages per-key rate limiters with LRU eviction.
 type RateLimiter struct {
-	mu sync.Mutex
+	// cacheMu protects the LRU cache which is not concurrency-safe.
+	cacheMu sync.Mutex
 
 	// configs is the list of rate limit configurations for this limiter.
 	configs []*RateLimitConfig
@@ -93,9 +94,6 @@ func NewRateLimiter(serviceName string, configs []*RateLimitConfig,
 // duration to wait if denied.
 func (rl *RateLimiter) Allow(r *http.Request, key string) (bool,
 	time.Duration) {
-
-	rl.mu.Lock()
-	defer rl.mu.Unlock()
 
 	path := r.URL.Path
 
@@ -183,9 +181,11 @@ func (rl *RateLimiter) Allow(r *http.Request, key string) (bool,
 }
 
 // getOrCreateLimiter retrieves an existing limiter or creates a new one.
-// Must be called with mu held.
 func (rl *RateLimiter) getOrCreateLimiter(key limiterKey,
 	cfg *RateLimitConfig) *rate.Limiter {
+
+	rl.cacheMu.Lock()
+	defer rl.cacheMu.Unlock()
 
 	// Try to get existing entry from cache (also updates LRU order).
 	if entry, err := rl.cache.Get(key); err == nil {
@@ -216,8 +216,8 @@ func (rl *RateLimiter) getOrCreateLimiter(key limiterKey,
 
 // Size returns the current number of entries in the cache.
 func (rl *RateLimiter) Size() int {
-	rl.mu.Lock()
-	defer rl.mu.Unlock()
+	rl.cacheMu.Lock()
+	defer rl.cacheMu.Unlock()
 
 	return rl.cache.Len()
 }


proxy/proxy.go
```diff
@@ -167,10 +168,47 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if res.OK() {
+			delay := res.Delay()
+			res.CancelAt(time.Now())
+			if delay > 0 {
+				// As seconds; for sub-second delays we still
+				// send 1 second.
+				secs := int(delay.Seconds())
+				if secs == 0 {
+					secs = 1
+				}
```
starius: We shouldn't floor the delay, since the client can try too early and get another 429 error.

I propose to round up when converting the delay to seconds (e.g., `secs := int(math.Ceil(delay.Seconds())`). Since we already know that `delay > 0` we can deduce that secs will be at least 1, so we don't have to enforce the minimum manually.

proxy/ratelimiter_integration_test.go
```diff
@@ -0,0 +1,347 @@
...
+
+	macBytes, err := dummyMac.MarshalBinary()
+	require.NoError(t, err)
+
+	macStr := base64.StdEncoding.EncodeToString(macBytes)
+
+	return fmt.Sprintf("L402 %s:%s", macStr, preimageHex)
+}
+
+func TestHTTPRateLimit_RetryAfterAndCORS(t *testing.T) {
```
starius: Add godoc

proxy/ratelimiter_integration_test.go
```diff
@@ -0,0 +1,347 @@
...
+	mockAuth := auth.NewMockAuthenticator()
+	p, err := proxy.New(mockAuth, services, []string{})
+	require.NoError(t, err)
+
+	srv := &http.Server{
+		Addr:    testProxyAddr,
+		Handler: http.HandlerFunc(p.ServeHTTP),
+	}
+	go func() { _ = srv.ListenAndServe() }()
+	t.Cleanup(func() { _ = srv.Close() })
```
starius: defer wait for all the goroutines to stop using a wait group.

proxy/ratelimiter_integration_test.go
```diff
@@ -0,0 +1,347 @@
...
+	err = resp.Body.Close()
+	require.NoError(t, err)
+
+	resp, err = client.Get(url)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+	require.Equal(t, http.StatusTooManyRequests, resp.StatusCode)
+}
+
+func TestGRPCRateLimit_ResponsesAndCORS(t *testing.T) {
```
starius: godoc

proxy/proxy.go
```diff
@@ -167,10 +167,49 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+		return
+	}
+
 	resourceName := target.ResourceName(r.URL.Path)
 
-	// Determine auth level required to access service and dispatch request
-	// accordingly.
+	// Determine the auth level required to access service and dispatch the
+	// request  accordingly.
```
starius: `s/t  a/t a/`

proxy/proxy.go
```diff
@@ -167,10 +168,47 @@ func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				if secs == 0 {
+					secs = 1
+				}
+				w.Header().Set(
+					"Retry-After", strconv.Itoa(secs),
+				)
+			}
+		}
+		addCorsHeaders(w.Header())
+		sendDirectResponse(
```
starius: `sendDirectResponse` always sets gRPC status to `codes.Internal`.

Shouldn't we use `codes.ResourceExhausted` in this situation so client side software understands that it can retry the same request later?

proxy/ratelimiter.go
```diff
@@ -0,0 +1,231 @@
...
+
+	for _, cfg := range rl.configs {
+		if !cfg.Matches(path) {
+			continue
+		}
+		anyMatched = true
+
+		// Create composite key: base key + path pattern for independent
+		// limiting per rule.
+		compositeKey := fmt.Sprintf("%s:%s", key, cfg.PathRegexp)
```
starius: IMHO for a map key, it is better to use a struct, not a concatenated string to save some RAM. `cfg.PathRegexp` will be repeated in many such concatenations, spending a separate memory for each key. In a struct a string field will be efficiently reused across many instances.

proxy/ratelimiter.go
```diff
@@ -0,0 +1,231 @@
...
+				bytes.NewBuffer(mac.Id()),
+			)
+			if err == nil {
+				return "token:" + identifier.TokenID.String()
+			}
+		}
+	}
+
+	// Fall back to IP address for unauthenticated requests.
+	return "ip:" + remoteIP.String()
```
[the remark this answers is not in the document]
hieblmi: I moved the ip masking to a package to `netutil`, and we key now on the ip range that a given request IP is mapping to.

starius (approved):
**LGTM!** 🏆
The code is great! The remaining issues are in the sample config and also one nit about rate limiter consistency when request is denied because of another (overlapping) rule.

netutil/ip_test.go
```diff
@@ -0,0 +1,109 @@
+package netutil
+
+import (
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMaskIP(t *testing.T) {
```
starius: add godoc

proxy/ratelimit_config.go
```diff
@@ -0,0 +1,53 @@
...
+	// compiledPathRegexp is the compiled version of PathRegexp.
+	compiledPathRegexp *regexp.Regexp
+}
+
+// Rate returns the rate.Limit value (requests per second) for this
+// configuration.
+func (r *RateLimitConfig) Rate() float64 {
+	if r.Per == 0 {
+		return 0
+	}
```
starius: add an empty line

proxy/ratelimit_config.go
```diff
@@ -0,0 +1,53 @@
...
+	return r.Burst
+}
+
+// Matches returns true if the given path matches this rate limit's path
+// pattern.
+func (r *RateLimitConfig) Matches(path string) bool {
+	if r.compiledPathRegexp == nil {
+		return true // No pattern means match all
+	}
```
starius: add an empty line

PR #232, opened 2026-04-15
aperture: support third-party caveat discharge macaroons

djkazic:
When a macaroon contains a third-party caveat, the macaroon library's `VerifySignature` walks the caveat chain and calls `findDischarge` to locate the corresponding discharge macaroon. With `nil` discharges, this search always fails, returning `"cannot find discharge macaroon for caveat ..."` which is propagated as a verification error. The library treats a missing discharge as a verification failure, not a skippable condition, so there was no way to verify such macaroons.

The fix uses `macaroon.Slice` (where the first element is the root macaroon and the rest are discharges) as the serialization format in both the base64 `Authorization` header and the hex-encoded `Grpc-Metadata-Macaroon`/`Macaroon` headers. `Slice.UnmarshalBinary` is a strict superset of `Macaroon.UnmarshalBinary` for single-macaroon inputs, so existing clients are unaffected.

### DIFF
--- a/l402/header.go
+++ b/l402/header.go
@@ -42,7 +42,13 @@ var (
 //
 // If only the macaroon is sent in header 2 or three then it is expected to have
 // a caveat with the preimage attached to it.
-func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, error) {
+//
+// The returned discharge macaroons will be non-nil when the binary-encoded
+// macaroon data contains more than one macaroon (as per the macaroon.Slice
+// convention, the first is the root macaroon and the rest are discharges for
+// its third-party caveats).
+func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage,
+	[]*macaroon.Macaroon, error) {
 	var authHeader string
 
 	switch {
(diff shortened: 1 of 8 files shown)


djkazic:
@claude review this

lightninglabs-deploy (approved):
Solid PR. The approach of using `macaroon.Slice` as the serialization format is clean and the backwards compatibility story is well thought out.

I verified that `Slice.MarshalBinary` for a single macaroon produces identical bytes to `Macaroon.MarshalBinary` (both use `appendBinary` under the hood), and `Slice.UnmarshalBinary` is indeed a strict superset of `Macaroon.UnmarshalBinary` — so existing single-macaroon clients are truly unaffected. 👍

A few observations/minor nits:

**1. `tokenFromContext` discards discharges**

In `l402/server_interceptor.go`, the new `_` blank identifier at the call site means that if anyone later needs the token ID from a request that also carries discharges, they'd have to refactor. This is fine for now since `tokenFromContext` only needs the token ID, but worth noting.

**2. `TestSetHeaderBackwardsCompatible` — consider also checking `Slice.MarshalBinary` directly**

The test currently round-trips through `SetHeader`/header parsing to check the base64 matches. It might be slightly more robust to also assert:
```
sliceBytes, _ := macaroon.Slice{mac}.MarshalBinary()
require.Equal(t, macBytes, sliceBytes)
```
This would make the invariant the test relies on explicit, rather than testing it indirectly through the header format string.

**3. `HasCaveat` only checks root macaroon caveats — correct**

Just confirming: `HasCaveat` iterates `m.Caveats()` which returns first-party caveats only on the root macaroon, so the preimage extraction in the hex-header path correctly ignores discharge macaroons. Good.

**4. Tests are thorough**

Nice coverage of:
- Auth header with/without discharges
- Hex header with/without discharges
- Multiple discharge round-trip
- Backwards compatibility
- Mint-level verification with/without discharges

The mint test correctly validates both the positive case (discharge present → verification succeeds) and the negative case (no discharge → `"cannot find discharge"` error).

Overall this is a clean, well-scoped change. LGTM.

l402/header.go
```diff
@@ -63,32 +69,50 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+					"failed: %v", err)
 		}
-		mac := &macaroon.Macaroon{}
-		err = mac.UnmarshalBinary(macBytes)
-		if err != nil {
-			return nil, lntypes.Preimage{}, fmt.Errorf("unable to "+
-				"unmarshal macaroon: %v", err)
+
+		// Use Slice to unmarshal so we can extract discharge
+		// macaroons if present. By convention the first macaroon
```
Roasbeef, line 87: Interesting, so now there'll be multiple macaroons, space delimited as part of the header? Do we need to update the spec to account for this?

l402/header.go
```diff
@@ -63,32 +69,50 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+		}
+		if len(slice) == 0 {
+			return nil, lntypes.Preimage{}, nil,
+				fmt.Errorf("no macaroon found in " +
+					"auth header")
+		}
+		mac := slice[0]
+		var discharges []*macaroon.Macaroon
+		if len(slice) > 1 {
+			discharges = slice[1:]
```
Roasbeef, line 103: Re spec again, right now depending on an implicit ordering here.

l402/header.go
```diff
@@ -99,43 +123,60 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
 	}
+	if len(slice) == 0 {
+		return nil, lntypes.Preimage{}, nil, fmt.Errorf(
+			"no macaroon found in header",
+		)
+	}
+	mac := slice[0]
+	var discharges []*macaroon.Macaroon
+	if len(slice) > 1 {
+		discharges = slice[1:]
```
Roasbeef, line 151: Same logic in a few areas, can extract into a helper functio.

l402/header.go
```diff
@@ -99,43 +123,60 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+	return mac, preimage, discharges, nil
 }
 
 // SetHeader sets the provided authentication elements as the default/standard
-// HTTP header for the L402 protocol.
+// HTTP header for the L402 protocol. If discharges are provided, they are
+// serialized alongside the root macaroon using the macaroon.Slice convention.
 func SetHeader(header *http.Header, mac *macaroon.Macaroon,
-	preimage fmt.Stringer) error {
+	preimage fmt.Stringer, discharges []*macaroon.Macaroon) error {
```
Roasbeef, line 173: Let's make the discharges ...*macaroon.Macaroon`, then we won't break the API.

l402/header.go
```diff
@@ -63,32 +69,50 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
 		}
 
 		if len(matches) != 4 {
-			return nil, lntypes.Preimage{}, fmt.Errorf("invalid "+
-				"auth header format: %s", authHeader)
+			return nil, lntypes.Preimage{}, nil,
```
Roasbeef, line 72: Slightly out of scope, but I realize there's a bug here: https://github.com/djkazic/aperture/blob/4e3194e169b577cc63b887331415a533043bab39/l402/header.go#L58-L69

That _should_ be a break.

l402/header.go
```diff
@@ -99,43 +123,60 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
-	mac := &macaroon.Macaroon{}
-	err = mac.UnmarshalBinary(macBytes)
-	if err != nil {
-		return nil, lntypes.Preimage{}, fmt.Errorf("unable to "+
+	var slice macaroon.Slice
+	if err := slice.UnmarshalBinary(macBytes); err != nil {
+		return nil, lntypes.Preimage{}, nil, fmt.Errorf("unable to "+
 			"unmarshal macaroon: %v", err)
 	}
+	if len(slice) == 0 {
```
Roasbeef, line 143: Should add a test case for this.

Roasbeef (approved):
LGTM 🐴

djkazic:
Will address review comments in the AM.

PR #241, opened 2026-06-10
Add L402 discovery: manifest + quote endpoints

Roasbeef:
In this PR, we add the server side of the L402 discovery layer to aperture, as
specified in [lightninglabs/L402#27](https://github.com/lightninglabs/L402/pull/27).
Today aperture is reactive: a client only learns what a service costs by hitting
it and getting a 402 back. Discovery lets a client see the catalog, the prices,
and the caveat vocabulary up front, and gives us a place for dynamic pricing and
the start of in-protocol negotiation.

Two new endpoints are mounted as priority local services, the same mechanism the
admin and dashboard endpoints use, so they bypass the 402 gate and are always
free. `/.well-known/l402.json` serves a static manifest built from the
configured services: their capabilities, tiers, prices, constraint bounds, and
the caveat conditions the proxy enforces. `/l402/quote` takes a client-chosen
bundle (service, capabilities, constraint values, optional budget), prices it,
maps it to the macaroon's first-party caveats, and mints a ready-to-pay
challenge.

## No new verification machinery

The quote endpoint reuses the existing mint and challenger. We factor the core
of `MintL402` into a shared helper and add `MintL402WithCaveats`, which takes an
explicit price and caveat set instead of deriving them from the static
`ServiceLimiter`. A quote response is just an ordinary L402 challenge minted for
a bundle the client chose, so the credential verifies through the unmodified
base path: nothing downstream of the mint changes.

## Formula pricing

We add a formula pricing model (`basemsat + sum(pricemsatperunit * ceil(value /
unit))`) as a value struct on `proxy.Service`, mirroring `DynamicPrice`. It is
surfaced only through discovery: the manifest advertises it, and the quote
endpoint computes the price from the client's chosen constraint values. The
reactive 402 price is untouched and still uses `Price` or `DynamicPrice`. The
sample config documents the new option.

The quote endpoint returns a structured error envelope on failure, sends
permissive CORS headers (the manifest carries no secrets), and the discovery
package ships with unit tests covering the manifest build, the pricing models,
the bundle-to-caveat mapping, and the error cases.

BOLT 12 is intentionally not wired here: a macaroon must commit to the payment
hash, which a BOLT 12 offer does not expose until an invoice is fetched, so that
binding needs its own treatment, as noted in the spec.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -31,6 +31,7 @@ import (
 	"github.com/lightninglabs/aperture/aperturedb"
 	"github.com/lightninglabs/aperture/auth"
 	"github.com/lightninglabs/aperture/challenger"
+	"github.com/lightninglabs/aperture/discovery"
 	"github.com/lightninglabs/aperture/lnc"
 	"github.com/lightninglabs/aperture/mint"
 	"github.com/lightninglabs/aperture/proxy"
@@ -1751,11 +1752,38 @@ func createProxy(cfg *Config, services []*proxy.Service,
 		},
 	))
 
+	// Wire the L402 discovery endpoints (the free manifest and the quote
+	// endpoint) as priority local services so they bypass the 402 gate, the
+	// same way the admin and dashboard endpoints do. They are always on, and
+	// reuse the same minter as the reactive flow so quote credentials are
+	// indistinguishable at verification time.
+	discoveryServer := discovery.New(discovery.Config{
+		Services: func() []*proxy.Service { return services },
+		Minter:   minter,
+		Provider: discovery.Provider{Name: providerName(cfg)},
+	})
+	priorityServices := append(
+		[]proxy.LocalService{
+			discoveryServer.ManifestService(),
+			discoveryServer.QuoteService(),
+		},
+		adminPriorityServices...,
+	)
+
 	prxy, err := proxy.New(
 		authenticator, services, cfg.Blocklist,
-		adminPriorityServices, localServices...,
+		priorityServices, localServices...,
 	)
 	return prxy, proxyCleanup, err
+}
+
+// providerName derives a human-readable provider name for the discovery
+// manifest from the configuration.
+func providerName(cfg *Config) string {
+	if cfg.Authenticator != nil && cfg.Authenticator.MPPRealm != "" {
+		return cfg.Authenticator.MPPRealm
+	}
+	return cfg.ListenAddr
 }
 
 // createHashMailServer creates the gRPC server for the hash mail message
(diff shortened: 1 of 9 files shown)


DrVelvetFog:
Nice to see the server side land — read through the discovery package and the mint refactor. One thing in the formula pricing path I'd flag before it ships, plus two smaller notes.

## Formula price can overflow int64 and underprice the quote

`FormulaConfig.PriceMsat` does unchecked int64 math on a client-supplied value:

```go
units := (value + unit - 1) / unit
price += c.PriceMsatPerUnit * units
```

The only upper-bound check in `resolveConstraints` is gated on the constraint being declared in `svc.Constraints`:

```go
if cfg, ok := svc.Constraints[cond]; ok {
    if max, err := ...; n > max { /* out of bounds */ }
}
```

A **formula-only** constraint (in `Formula.Components` but not also in `svc.Constraints`) is in the `allowed` set but skips that bound. Sent as a JSON string it survives to an exact int64 via `stringifyValue` → `ParseInt`, so the multiplication can wrap.

Concretely: a `1000` msat/unit component (unit 1) with `value = 9223372036854775807` makes `PriceMsat` wrap to `0`. Then `ceilDiv(0,1000)=0` → `priceSats=0` → `priceMsat=0`, which passes the `max_price_msat` guard (`0 > max` is false) and goes to `MintL402WithCaveats` as the invoice amount. So a client picks a huge quantity and pays ~nothing.

Two things stack here:
1. formula components aren't required to map to a bounded constraint, so a no-upper-bound config exists;
2. even bounded constraints use unchecked multiplication — a large in-bounds `max` × large `PriceMsatPerUnit` can still overflow.

The reactive 402 path never had this exposure since `price` came from the operator's `ServiceLimiter`; the quote endpoint is the first place a client value reaches the mint, so a non-positive/wrapped amount is new territory for the challenger.

Suggested: require each formula component's constraint to resolve to a bounded one (checked at config load), do the price math with checked/saturating arithmetic, and reject a non-positive final price before minting.

## Smaller

- `priceBundle` computes the formula in msat then `ceilDiv`s to whole sats, and `processQuote` multiplies back — so `price_msat` in the response is sat-rounded, not what the manifest formula gives. A client pricing from the manifest (doc says it's "computable by a client from the manifest alone") gets e.g. 1500 where the quote says 2000. Probably worth a line that quotes round up to whole sats.
- negative constraint values are silently dropped in `PriceMsat` (`value <= 0` → skip) rather than 400'd; almost certainly a client bug worth rejecting.

Happy to be wrong if I've misread the bounds flow — that's the part I traced hardest.

PR #245, opened 2026-07-01
l402: harden authorization header parsing

grandpig:
## Summary

- Harden parsing of L402 authorization headers.
- Reject empty, duplicate, malformed, or extra authentication headers.
- Accept one LSAT, one L402, or matching LSAT and L402 credentials in either order.
- Deduplicate macaroon and preimage decoding across supported header formats.

## Details

`SetHeader` emits both LSAT and L402 credentials for compatibility with older Aperture versions. When both are present, `FromHeader` requires their encoded macaroon and encoded preimage values to be identical.

Aperture currently supports exactly one macaroon per credential. The comma-separated multiple-macaroon form allowed by the L402 specification is not supported.

An Authorization header is considered present even when its value is empty.

This prevents an empty Authorization value from bypassing validation and falling back to another macaroon header.

## Testing

- `go test ./auth ./l402 -count=1`
- `go test ./...`

### DIFF
--- a/auth/authenticator_test.go
+++ b/auth/authenticator_test.go
@@ -132,6 +132,17 @@ func TestL402Authenticator(t *testing.T) {
 				result: true,
 			},
 			{
+				id: "valid auth header followed by unrelated value",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+						"Bearer unrelated",
+					},
+				},
+				result: true,
+			},
+			{
 				id: "valid macaroon metadata header",
 				header: &http.Header{
 					l402.HeaderMacaroonMD: []string{
--- a/auth/authenticator_test.go
+++ b/auth/authenticator_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/lightninglabs/aperture/auth"
@@ -14,8 +15,13 @@ import (
 
 // createDummyMacHex creates a valid macaroon with dummy content for our tests.
 func createDummyMacHex(preimage string) string {
+	return createDummyMacHexWithID(preimage, "AA==")
+}
+
+// createDummyMacHexWithID creates a valid macaroon with the given ID.
+func createDummyMacHexWithID(preimage, id string) string {
 	dummyMac, err := macaroon.New(
-		[]byte("aabbccddeeff00112233445566778899"), []byte("AA=="),
+		[]byte("aabbccddeeff00112233445566778899"), []byte(id),
 		"aperture", macaroon.LatestVersion,
 	)
 	if err != nil {
@@ -33,16 +39,43 @@ func createDummyMacHex(preimage string) string {
 	return hex.EncodeToString(macBytes)
 }
 
+// createDummyMacHexWithoutPreimage creates a valid macaroon without a
+// preimage caveat.
+func createDummyMacHexWithoutPreimage() string {
+	dummyMac, err := macaroon.New(
+		[]byte("aabbccddeeff00112233445566778899"), []byte("AA=="),
+		"aperture", macaroon.LatestVersion,
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	macBytes, err := dummyMac.MarshalBinary()
+	if err != nil {
+		panic(err)
+	}
+
+	return hex.EncodeToString(macBytes)
+}
+
 // TestL402Authenticator tests that the authenticator properly handles auth
 // headers and the tokens contained in them.
 func TestL402Authenticator(t *testing.T) {
 	var (
 		testPreimage = "49349dfea4abed3cd14f6d356afa83de" +
 			"9787b609f088c8df09bacc7b4bd21b39"
-		testMacHex      = createDummyMacHex(testPreimage)
-		testMacBytes, _ = hex.DecodeString(testMacHex)
-		testMacBase64   = base64.StdEncoding.EncodeToString(
+		otherPreimage = "59349dfea4abed3cd14f6d356afa83de" +
+			"9787b609f088c8df09bacc7b4bd21b39"
+		testMacHex       = createDummyMacHex(testPreimage)
+		noPreimageMacHex = createDummyMacHexWithoutPreimage()
+		otherIDMacHex    = createDummyMacHexWithID(testPreimage, "BB==")
+		testMacBytes, _  = hex.DecodeString(testMacHex)
+		testMacBase64    = base64.StdEncoding.EncodeToString(
 			testMacBytes,
+		)
+		otherIDMacBytes, _ = hex.DecodeString(otherIDMacHex)
+		otherIDMacBase64   = base64.StdEncoding.EncodeToString(
+			otherIDMacBytes,
 		)
 		headerTests = []struct {
 			id       string
@@ -100,12 +133,133 @@ func TestL402Authenticator(t *testing.T) {
 				result: false,
 			},
 			{
+				id: "macaroon header without preimage caveat",
+				header: &http.Header{
+					l402.HeaderMacaroon: []string{noPreimageMacHex},
+				},
+				result: false,
+			},
+			{
+				id: "empty auth header with valid fallback macaroon",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{""},
+					l402.HeaderMacaroon:      []string{testMacHex},
+				},
+				result: false,
+			},
+			{
+				id: "empty metadata header with valid macaroon",
+				header: &http.Header{
+					l402.HeaderMacaroonMD: []string{""},
+					l402.HeaderMacaroon:   []string{testMacHex},
+				},
+				result: false,
+			},
+			{
+				id: "multiple macaroon metadata headers",
+				header: &http.Header{
+					l402.HeaderMacaroonMD: []string{
+						testMacHex, testMacHex,
+					},
+				},
+				result: false,
+			},
+			{
+				id: "multiple macaroon headers",
+				header: &http.Header{
+					l402.HeaderMacaroon: []string{
+						testMacHex, testMacHex,
+					},
+				},
+				result: false,
+			},
+			{
 				id: "valid auth header (both LSAT and L402)",
 				header: &http.Header{
 					l402.HeaderAuthorization: []string{
 						"LSAT " + testMacBase64 + ":" +
 							testPreimage,
 						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+					},
+				},
+				result: true,
+			},
+			{
+				id: "valid auth header (both L402 and LSAT)",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+					},
+				},
+				result: true,
+			},
+			{
+				id: "different macaroons (LSAT then L402)",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+						"L402 " + otherIDMacBase64 + ":" +
+							testPreimage,
+					},
+				},
+				result: false,
+			},
+			{
+				id: "different macaroons (L402 then LSAT)",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 " + otherIDMacBase64 + ":" +
+							testPreimage,
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+					},
+				},
+				result: false,
+			},
+			{
+				id: "different preimages (LSAT then L402)",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+						"L402 " + testMacBase64 + ":" +
+							otherPreimage,
+					},
+				},
+				result: false,
+			},
+			{
+				id: "different preimages (L402 then LSAT)",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 " + testMacBase64 + ":" +
+							otherPreimage,
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+					},
+				},
+				result: false,
+			},
+			{
+				id: "case insensitive scheme and uppercase preimage",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"lSaT " + testMacBase64 + ":" +
+							strings.ToUpper(testPreimage),
+					},
+				},
+				result: true,
+			},
+			{
+				id: "multiple spaces after scheme",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402   " + testMacBase64 + ":" +
 							testPreimage,
 					},
 				},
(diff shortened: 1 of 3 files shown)


grandpig:
@starius Could you please take a look when you have a chance? Thanks!

auth/authenticator_test.go
```diff
@@ -131,6 +131,17 @@ func TestL402Authenticator(t *testing.T) {
 				},
 				result: true,
 			},
+			{
+				id: "valid auth header followed by unrelated value",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+						"Bearer unrelated",
```
starius, line 294: Another case is broken:
```
Authorization: L402 !!!!:<64-hex-preimage>
Authorization: L402 <valid-macaroon>:<valid-preimage>
```

The first value matches the permissive regex, the loop stops, and decoding fails. Reversing the values succeeds. The parent implementation would use the final valid match, so this is a newly introduced regression.

l402/header.go
```diff
@@ -60,6 +60,8 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
 			if len(matches) != 4 {
 				continue
 			}
+
+			break
```
starius: Could you provide the context of this change, please?

In general, [the L402 protocol](https://github.com/lightninglabs/L402/blob/master/protocol-specification.md#52-credentials-authorization) does not require implementations to support multiple `Authorization: L402` headers.

Therefore, accepting an L402 among unrelated repeated values is a compatibility extension, not L402 compliance.

starius (changes requested):

litbot-9000:
@grandpig, remember to re-request review from reviewers when ready

auth/authenticator_test.go
grandpig: Done

l402/header.go
grandpig: Thanks. Modified.

grandpig:
> @grandpig, remember to re-request review from reviewers when ready

@starius @starius 

Thanks, good catch. The break introduced a regression when an earlier Authorization value matches the permissive regex but contains an invalid macaroon, while a later value is valid.

I changed the loop to retain the last syntactically matching LSAT/L402 value while ignoring unrelated Authorization values, preserving the parent implementation's behavior. I also added a regression test for
the invalid-then-valid ordering.

The multi-value compatibility is needed because SetHeader currently emits both LSAT and L402 Authorization values to support older Aperture clients. Supporting unrelated values such as Bearer is an additional compatibility extension, and I added coverage for that behavior as well.

starius:
Hmm, so now it permits requests like this:
```
Authorization: L402 <valid>
Authorization: Bearer unrelated 
```

Why are they needed? Note that requests like
```
Authorization: L402 <valid>
Authorization: LSAT <valid> 
```
are already supported.

I'm worried that permitting non-conformant requests like ones adding `Authorization: Bearer unrelated` to a valid LSAT/L402 increases attack surface and must be avoided.

I would go in the opposite direction: reject requests with valid-after-invalid headers. This looks a valid hardening direction to me. What do you think?

I.e. permit exactly one of these forms:
 - valid LSAT
 - valid L402
 - valid LSAT + valid L402
 - valid L402 + valid LSAT
 - anything else in `Authorization` header => reject

Thoughts?

grandpig:
> Thoughts?


Agreed. I don't have a concrete compatibility requirement for accepting an unrelated Bearer value, and I agree that allowing it would broaden the accepted request surface unnecessarily.

I tightened FromHeader so that every Authorization value must be a valid LSAT or L402 credential. It now accepts exactly one LSAT, one L402, or one of each in either order. Any malformed, unrecognized, duplicate, or extra Authorization value is rejected, including both valid-before-invalid and invalid-before-valid cases.

I also added regression tests covering these cases.

grandpig:
@starius Modified. Please review it again. Thansk

starius:
@grandpig 

There are three blocking gaps.

1. At `l402/header.go:99`, each parsed credential overwrites the previous one. Only the last credential is cryptographically verified by `authenticator.go:44`.

Therefore this still succeeds:

```text
Authorization: LSAT <well-formed but cryptographically invalid macaroon>:<preimage>
Authorization: L402 <valid credential>
```

The reverse order fails. So cryptographic invalid-before-valid remains order-dependent.

The test uses `!!!!`, which fails Base64 decoding. It proves malformed input is rejected, but not that a structurally valid yet unauthorized credential is rejected. The mock minter also accepts every credential.

2. The LSAT/L402 exception should require both decoded macaroon bytes and preimages to be identical. Then validating either copy validates both. Add tests with differing but well-formed credentials in both orders.

3. Also, the new anchored regex is an improvement, but it still rejects protocol-valid variants: authentication schemes are case-insensitive, L402 permits `1*SP`, and hex can contain uppercase digits. See the [L402 grammar](https://github.com/lightninglabs/L402/blob/master/protocol-specification.md#53-grammar).

grandpig:
> @grandpig
> 
> There are three blocking gaps.
> 
> 1. At `l402/header.go:99`, each parsed credential overwrites the previous one. Only the last credential is cryptographically verified by `authenticator.go:44`.
> 
> Therefore this still succeeds:
> 
> ```
> Authorization: LSAT <well-formed but cryptographically invalid macaroon>:<preimage>
> Authorization: L402 <valid credential>
> ```
> 
> The reverse order fails. So cryptographic invalid-before-valid remains order-dependent.
> 
> The test uses `!!!!`, which fails Base64 decoding. It proves malformed input is rejected, but not that a structurally valid yet unauthorized credential is rejected. The mock minter also accepts every credential.
> 
> 2. The LSAT/L402 exception should require both decoded macaroon bytes and preimages to be identical. Then validating either copy validates both. Add tests with differing but well-formed credentials in both orders.
> 3. Also, the new anchored regex is an improvement, but it still rejects protocol-valid variants: authentication schemes are case-insensitive, L402 permits `1*SP`, and hex can contain uppercase digits. See the [L402 grammar](https://github.com/lightninglabs/L402/blob/master/protocol-specification.md#53-grammar).

Thanks, these were good catches.

I updated FromHeader so that when both LSAT and L402 are present, their decoded macaroon bytes and decoded preimages must be identical. A credential is no longer allowed to overwrite a previous credential, so mismatched credentials are rejected in either order.

I also added tests using structurally valid but cryptographically unauthorized macaroons, including both invalid-before-valid and valid-before-invalid ordering.

Finally, I relaxed the parser according to the L402 grammar: authentication schemes are case-insensitive, 1*SP is accepted, and uppercase hexadecimal preimages are accepted.

l402/header.go
```diff
@@ -49,41 +54,64 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+				return nil, lntypes.Preimage{}, fmt.Errorf("invalid "+
+					"auth header format: %s", authHeader)
 			}
-		}
 
-		if len(matches) != 4 {
-			return nil, lntypes.Preimage{}, fmt.Errorf("invalid "+
-				"auth header format: %s", authHeader)
-		}
+			scheme := strings.ToUpper(matches[1])
```
starius: Let's unpack here so we could use nice variable names below, not `matches[x]`:
```go
scheme, macBase64, preimageHex := strings.ToUpper(matches[1]), matches[2], matches[3]
```

auth/authenticator_test.go
```diff
@@ -131,6 +224,87 @@ func TestL402Authenticator(t *testing.T) {
...
+					l402.HeaderAuthorization: []string{
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+						"Bearer unrelated",
+					},
+				},
+				result: false,
+			},
```
starius, line 467: I found gaps in test coverage:

1. Base64 decoding failure at `header.go:85`.

   The existing `!!!!` tests fail the regex first, so they never reach `DecodeString`. A regex-compatible invalid value such as `A` would exercise this.

2. Macaroon unmarshalling failure at `header.go:92`.

   This needs valid Base64 containing invalid macaroon bytes, for example `YQ==`.

3. Preimage conversion failure at `header.go:100`.

   This path is effectively unreachable: the regex already guarantees exactly 64 hexadecimal characters, which satisfies every validation performed by `MakePreimageFromStr`.

Please test for Base64-valid/non-macaroon data in both positions:

```
invalid encoded macaroon → valid credential
valid credential → invalid encoded macaroon
```

l402/header.go
```diff
@@ -49,41 +54,64 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+			if mac != nil && (!bytes.Equal(macBytes, currentMacBytes) ||
+				preimage != currentPreimage) {
+				return nil, lntypes.Preimage{}, errors.New(
+					"authorization credentials do not match",
+				)
+			}
+
+			mac = currentMac
+			macBytes = currentMacBytes
+			preimage = currentPreimage
```
starius:
```go
if mac == nil {
        mac = currentMac
        macBytes = currentMacBytes
        preimage = currentPreimage

} else if !bytes.Equal(macBytes, currentMacBytes) ||
        preimage != currentPreimage:

        return nil, lntypes.Preimage{}, errors.New(
                "authorization credentials do not match",
        )
}
```

starius (commented):
The changes look good. It is a reasonable hardening. Here are some nits in the code. Please fix and squash all the commits and sync PR title and description to the final shape of the PR.

l402/header.go
```diff
@@ -27,7 +29,10 @@ const (
 )
 
 var (
-	authRegex        = regexp.MustCompile("(LSAT|L402) (.*?):([a-f0-9]{64})")
+	authRegex = regexp.MustCompile(
```
starius, line 38: Let's provide the format that we look for in the godoc of this reghexp, we can take it from the spec: https://github.com/lightninglabs/L402/blob/master/protocol-specification.md#52-credentials-authorization

Also I found that the spec supports multiple macaroons there, separated by a comma: "Multiple macaroons are base64-encoded individually and comma-separated before the colon." Aperture does not support this currently, AFAIK, and also I am not aware about a use case for multiple macaroons. I propose just to document in godoc that this multiple macaroons are not supported now, despite being allowed by the spec.

l402/header.go
grandpig: Done

auth/authenticator_test.go
grandpig: Done

l402/header.go
grandpig: Done

l402/header.go
grandpig: Done

grandpig:
> The changes look good. It is a reasonable hardening. Here are some nits in the code. Please fix and squash all the commits and sync PR title and description to the final shape of the PR.

`SetHeader` emits matching LSAT and L402 credentials for compatibility with older Aperture versions. When both are present, `FromHeader` now requires their decoded macaroon bytes and preimages to be identical. This ensures that
cryptographically validating either credential validates both copies.

Please review it again. Thnaks.

l402/header.go
```diff
@@ -49,41 +61,66 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
 	// Header field 1 contains the macaroon and the preimage as distinct
 	// values separated by a colon.
 	case header.Get(HeaderAuthorization) != "":
```
starius: Preexisting problem here. Not an authentication bypass, but still worth hardening.

`Header.Get` returns the first value, and returns `""` both when the header is absent and when its first value is empty. Consequently:

```http
Authorization:
Macaroon: <valid legacy macaroon>
```

skips Authorization validation and authenticates through the fallback `Macaroon` header. That contradicts the stated rule that every supplied Authorization value must be valid. It also remains order-dependent when the first Authorization value is empty.

Use presence rather than non-empty content:

```go
authHeaders := header.Values(HeaderAuthorization)

switch {
case len(authHeaders) > 0:
	for _, authHeader := range authHeaders {
		// An empty value now reaches the regex and is rejected.
	}
```

Add a regression case containing an empty Authorization value plus an otherwise valid fallback macaroon and expect rejection.

---

Two other cases should also be fixed:

```go
  case len(macaroonMDHeaders) > 0:
        if len(macaroonMDHeaders) != 1 {
                return nil, lntypes.Preimage{}, errors.New(
                        "multiple macaroon metadata headers",
                )
        }
        authHeader = macaroonMDHeaders[0]

  case len(macaroonHeaders) > 0:
        if len(macaroonHeaders) != 1 {
                return nil, lntypes.Preimage{}, errors.New(
                        "multiple macaroon headers",
                )
        }
        authHeader = macaroonHeaders[0]
```

to reject requests like:
```
Grpc-Metadata-Macaroon:
Macaroon: <valid credential>
```

l402/header.go
```diff
@@ -49,41 +61,66 @@ func FromHeader(header *http.Header) (*macaroon.Macaroon, lntypes.Preimage, erro
...
+				mac = currentMac
+				macBytes = currentMacBytes
+				preimage = currentPreimage
+			} else if !bytes.Equal(macBytes, currentMacBytes) ||
+				preimage != currentPreimage {
+
+				return nil, lntypes.Preimage{}, errors.New(
+					"authorization credentials do not match",
+				)
+			}
```
starius, line 103: Look at the code after the end of the `switch` - it repeats most of the logic that we have for Authorization headers:

  - macaroon.Macaroon.UnmarshalBinary                                                                                                                                                                                                                         
  - lntypes.MakePreimageFromStr                                                                                                                                                                                                                               
  - the corresponding error handling                                                                                                                                                                                                                          
  - returning the final macaroon, preimage

It only gets the preimage another way (`HasCaveat(mac, PreimageKey)`).

Would you like to deduplicate this? `Authorization` case could only check LSAT and L402 equality and fill the variables, not do full decoding and validation. It will be more strict since it will compare encoded forms, not decoded, but this is fine. 

Conceptually:

```go
var (                                                                                                                                                                                                                                                       
        macBytes    []byte                                                                                                                                                                                                                                    
        preimageHex string                                                                                                                                                                                                                                    
)

switch {
case len(authHeaders) > 0:
  // Check the regexp and equality of encoded forms if both LSAT and L402 are provided, fill macBase64 and preimageHex

  case len(macaroonMDHeaders) > 0:
        // Check macaroonMDHeaders length and fill macBase64 and Hex-decode into macBytes.

  case len(macaroonHeaders) > 0:
        // Check macaroonHeaders length and fill macBase64 and Hex-decode into macBytes.
}

// Work with macBytes and preimageHex
// return  mac, preimage
```

PR description also needs to be adjusted to mention that encoded forms are compared.

starius (commented):
Thank you for updates!

I found another hardening vector in this function and also a way to deduplicate some code. Could you fix it, while we are on this, please?

grandpig:
> Thank you for updates!
> 
> I found another hardening vector in this function and also a way to deduplicate some code. Could you fix it, while we are on this, please?

Thanks, fixed.

FromHeader now uses header.Values to distinguish an absent header from an empty value. Any supplied empty Authorization value is rejected instead of falling back to Macaroon or Grpc-Metadata-Macaroon.

Multiple metadata macaroon headers and multiple Macaroon headers are also rejected.

I deduplicated the decoding and validation logic across all supported header formats. When both LSAT and L402 are present, their encoded macaroon and encoded preimage values must match exactly.

Please review it again.

l402/header.go
```diff
@@ -43,84 +54,110 @@ var (
...
 	switch {
 	// Header field 1 contains the macaroon and the preimage as distinct
 	// values separated by a colon.
-	case header.Get(HeaderAuthorization) != "":
-		// Parse the content of the header field and check that it is in
-		// the correct format.
-		var matches []string
-		authHeaders := header.Values(HeaderAuthorization)
+	case len(authHeaders) > 0:
+		var scheme string
```
starius: Duplicate second scheme is accepted. `var scheme` remembers only the first scheme. It therefore accepts:

```http
Authorization: L402 M:P
Authorization: LSAT M:P
Authorization: LSAT M:P
```

The third value is compared only against the first scheme (`L402`), so it is not detected as a duplicate. The inverse also works:

```http
Authorization: LSAT M:P
Authorization: L402 M:P
Authorization: L402 M:P
```

This contradicts "exactly one LSAT, one L402, or one of each." Use a `seenSchemes` map or reject `len(authHeaders) > 2`.

Add both three-value cases as tests:

```diff
diff --git a/auth/authenticator_test.go b/auth/authenticator_test.go
index 48886da..0d8925b 100644
--- a/auth/authenticator_test.go
+++ b/auth/authenticator_test.go
@@ -379,6 +379,34 @@ func TestL402Authenticator(t *testing.T) {
 				},
 				result: false,
 			},
+			{
+				id: "duplicate LSAT after matching L402 and LSAT",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+					},
+				},
+				result: false,
+			},
+			{
+				id: "duplicate L402 after matching LSAT and L402",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"LSAT " + testMacBase64 + ":" +
+							testPreimage,
+						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+					},
+				},
+				result: false,
+			},
 			{
 				id: "extra auth header value",
 				header: &http.Header{
```

auth/authenticator_test.go
```diff
@@ -131,6 +258,140 @@ func TestL402Authenticator(t *testing.T) {
...
+				id: "valid credential followed by invalid encoded macaroon",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 " + testMacBase64 + ":" +
+							testPreimage,
+						"LSAT YQ==:" + testPreimage,
+					},
+				},
+				result: false,
+			},
```
starius, line 393: These test cases now fail at encoded credential comparison:
```go
  currentMacBase64 != macBase64
```
They never reach Base64 decoding or Authorization macaroon unmarshalling.

Coverage confirms:

- `FromHeader`: 96.1%
- Base64 decoder error at `header.go:138`: uncovered
- Missing-preimage-caveat error: uncovered, but pre-existing
- Macaroon unmarshal block is covered through the empty metadata-header case, not through `Authorization: L402 YQ==:...`

Add single-header cases:

```text
L402 A:<valid-preimage>
L402 YQ==:<valid-preimage>
```

```diff
diff --git a/auth/authenticator_test.go b/auth/authenticator_test.go
index 0d8925b..9b23dbb 100644
--- a/auth/authenticator_test.go
+++ b/auth/authenticator_test.go
@@ -39,6 +39,25 @@ func createDummyMacHexWithID(preimage, id string) string {
 	return hex.EncodeToString(macBytes)
 }
 
+// createDummyMacHexWithoutPreimage creates a valid macaroon without a
+// preimage caveat.
+func createDummyMacHexWithoutPreimage() string {
+	dummyMac, err := macaroon.New(
+		[]byte("aabbccddeeff00112233445566778899"), []byte("AA=="),
+		"aperture", macaroon.LatestVersion,
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	macBytes, err := dummyMac.MarshalBinary()
+	if err != nil {
+		panic(err)
+	}
+
+	return hex.EncodeToString(macBytes)
+}
+
 // TestL402Authenticator tests that the authenticator properly handles auth
 // headers and the tokens contained in them.
 func TestL402Authenticator(t *testing.T) {
@@ -47,10 +66,11 @@ func TestL402Authenticator(t *testing.T) {
 			"9787b609f088c8df09bacc7b4bd21b39"
 		otherPreimage = "59349dfea4abed3cd14f6d356afa83de" +
 			"9787b609f088c8df09bacc7b4bd21b39"
-		testMacHex      = createDummyMacHex(testPreimage)
-		otherIDMacHex   = createDummyMacHexWithID(testPreimage, "BB==")
-		testMacBytes, _ = hex.DecodeString(testMacHex)
-		testMacBase64   = base64.StdEncoding.EncodeToString(
+		testMacHex       = createDummyMacHex(testPreimage)
+		noPreimageMacHex = createDummyMacHexWithoutPreimage()
+		otherIDMacHex    = createDummyMacHexWithID(testPreimage, "BB==")
+		testMacBytes, _  = hex.DecodeString(testMacHex)
+		testMacBase64    = base64.StdEncoding.EncodeToString(
 			testMacBytes,
 		)
 		otherIDMacBytes, _ = hex.DecodeString(otherIDMacHex)
@@ -112,6 +132,13 @@ func TestL402Authenticator(t *testing.T) {
 				},
 				result: false,
 			},
+			{
+				id: "macaroon header without preimage caveat",
+				header: &http.Header{
+					l402.HeaderMacaroon: []string{noPreimageMacHex},
+				},
+				result: false,
+			},
 			{
 				id: "empty auth header with valid fallback macaroon",
 				header: &http.Header{
@@ -302,6 +329,24 @@ func TestL402Authenticator(t *testing.T) {
 				},
 				result: false,
 			},
+			{
+				id: "invalid base64 credential",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 A:" + testPreimage,
+					},
+				},
+				result: false,
+			},
+			{
+				id: "invalid encoded macaroon credential",
+				header: &http.Header{
+					l402.HeaderAuthorization: []string{
+						"L402 YQ==:" + testPreimage,
+					},
+				},
+				result: false,
+			},
 			{
 				id: "invalid base64 followed by valid credential",
 				header: &http.Header{

```

starius (commented):
Found a bug and a test coverage gap.

Also could you update the PR title?
"auth: preserve valid authorization header matches" -> "l402: harden authorization header parsing"

litbot-9000:
@grandpig, remember to re-request review from reviewers when ready

auth/authenticator_test.go
grandpig: Done

grandpig:
> **[litbot-9000](/litbot-9000) **

@litbot-9000 Please review it again.

l402/header.go
```diff
@@ -43,84 +54,111 @@ var (
...
+				preimageHex = currentPreimageHex
 				continue
 			}
-		}
 
-		if len(matches) != 4 {
-			return nil, lntypes.Preimage{}, fmt.Errorf("invalid "+
-				"auth header format: %s", authHeader)
+			if currentMacBase64 != macBase64 ||
+				currentPreimageHex != preimageHex {
```
starius, line 99: If preimages are the same, but encoded differently (lowercase vs uppercase), they are considered different here.

We can easily fix this if we apply lowercasing at line 84:
```go
currentPreimageHex := strings.ToLower(matches[3])
```

Add a regression case containing matching LSAT/L402 credentials where one preimage is uppercase and the other lowercase. (Matching LSAT/L402 credentials containing lowercase versus uppercase preimages.)

```
     LSAT M:abcdef...
     L402 M:ABCDEF...
```

starius (approved):
**LGTM!** 🎉

Found one remaining issue.

PR #247, opened 2026-07-06
pricesrpc+proxy: metered draw-down of prepaid L402 tokens, with a reference pricer daemon

Roasbeef:
In this PR, we extend the dynamic pricing hook so an external pricer can meter prepaid L402 tokens over their whole lifecycle, rather than only quoting a price at challenge time. The motivating shape is any metered, streaming service where the true cost of a request is only known after the response completes: LLM inference is the canonical example, where a client buys a bundle of tokens with one payment, draws it down across many streaming calls, and should be re-challenged exactly when the bundle runs dry. Today aperture only consults the pricer on unauthenticated requests, so once a token is minted it grants access until expiry and the seller has no per-request say and no view of actual usage.

## The new pricesrpc surface

`GetPrice` stays as is. Three RPCs join it, all invoked by aperture against the configured price server:

- **ChallengeMinted**: fired before a fresh 402 challenge goes out, carrying the minted macaroon's token ID and the quoted price. This is the association step: once the client pays, the pricer knows exactly which purchased balance that token anchors. If the notification fails, the challenge is aborted, since the client would otherwise pay for a bundle the pricer will not honor.
- **AuthorizeRequest**: invoked on every authenticated request to a metered service. The pricer answers allow, or deny with the price for the next challenge, and aperture mints a fresh 402 on denial. This closes the token-reuse gap for metered services: one payment no longer means unlimited access until expiry.
- **ReportUsage**: invoked once the response body completes (or the client disconnects), carrying a capped tail of the body. For SSE streams the trailing chunks include the final usage object, which is where the actual cost lives; the pricer does the cost analysis and debits the balance.

Metering is opt-in per service via `dynamicprice.metered`, with `usagetailbytes` bounding the captured tail. Requests authenticated through non-L402 schemes (e.g. MPP sessions) skip pricer metering, as session draw-down is already accounted for by the session authenticator, so the two models compose.

## Proxy integration

The authorize check runs right after successful authentication; the response observer wraps the body in the `ModifyResponse` hook, captures a bounded tail as bytes stream through, and reports usage exactly once on EOF or early close, off the request path. The copy loop itself is untouched, so SSE pass-through behavior (immediate flushing) does not change.

While here, two small fixes the work surfaced: request serialization for the pricer now uses `httputil.DumpRequest`, which restores the body it reads (the old `r.Write` consumed it outright, so a zero-price dynamic response forwarded a bodyless request to the backend), and the freebie path now sends the dynamically fetched price into the challenge instead of the static service price.

## meterd: a reference metering pricer

`meterd` (new `cmd/meterd`) is a small daemon implementing the full service: it quotes token bundles at a blended per-token msat rate from a per-model price table, books balances on ChallengeMinted, draws them down from the usage objects it extracts out of SSE or JSON response tails (front-truncation safe), and denies AuthorizeRequest with a re-quote once a bundle is spent. State persists as JSON across restarts. It doubles as the executable spec for what a metering price server must implement.

The full cycle has been exercised end to end on regtest against a real LND backend: 402 quoted and booked, invoice paid over a channel, paid retry authorized and streamed, exact usage debited, token reused without re-payment, and a drained bundle answered with a fresh challenge.

See each commit message for a detailed description w.r.t the incremental changes.


Roasbeef:
Addressed the review findings and force-pushed (branch `inference-pricing`, 4 commits, autosquashed). `go build`/`go vet`/`gofmt` clean; `meterd`, `pricer`, and the metering proxy tests all pass.

**Metering bypasses closed (must-fix):**
- **gzip evasion → no more free inference.** Metered requests now strip the client's `Accept-Encoding`, so the upstream response comes back plaintext (Go's transport transparently decompresses) and the usage tail is always parseable. As defense-in-depth, `Content-Encoding` now rides the `ReportUsage` proto so a non-identity encoding is a flagged error rather than a silent zero-debit. New test: a request with `Accept-Encoding: gzip` still yields a correct debit.
- **Unauthenticated challenge spam.** Bundle-state persistence is now coalesced (dirty flag + periodic flush + shutdown drain) instead of rewriting the whole state file per booking, and never-authorized bundles are capped with oldest-first eviction, so mint-spam can't grow memory unbounded or amplify disk I/O.

**Correctness / accounting (should-fix):**
- **TOCTOU overdraw.** The store now reserves an estimated per-request cost at `AuthorizeRequest` (netted out of the available balance), so N concurrent requests on a near-empty bundle can't all authorize; the reservation is released on `ReportUsage`.
- **Model substitution.** `AuthorizeRequest` now denies when the request's model differs from the bundle's booked model, quoting a fresh challenge for the model actually requested — a cheap-model bundle can't be spent on an expensive model.
- **Report-loss.** `ReportUsage` now retries with exponential backoff before giving up, and a terminal failure logs loudly (a durable un-acked-report queue is noted as the real follow-up).
- **Hot-path body buffering.** The pricer RPCs now serialize only a bounded 64 KiB prefix of the request body (enough for the model field), restoring the full body for the backend — no more full-body copy on every authenticated request or 4 MiB gRPC-limit 500s on large uploads.
- **Config validation.** `metered` now requires `enabled` (startup error), `UsageTailBytes` is bounded, and the msat price arithmetic has an overflow guard.

**Nits:** a metered request whose L402 header is present-but-unparseable now returns 500 instead of free unmetered access; the usage brace-matcher requires `"usage"` in JSON-key position so an incidental substring can't latch a wrong object.

**Note on CI vs local:** the pre-existing `TestProxyHTTP`/`TestProxyGRPC`/`TestProxyMPP` tests bind fixed ports and do a gRPC DNS TXT lookup; in my sandbox they fail identically on clean `master` (502 / DNS hang), unrelated to this change. The metering-specific tests and the changed-logic packages (`meterd`, `pricer`) all pass.

Roasbeef:
Pushed one more commit, da3b973: meterd now exposes a `Store` interface (the JSON file store becomes `JSONStore`, its default implementation) and a `RateSource` interface owning model resolution plus per-model bundle sizing, with `WithStore`/`WithRateSource` options on `NewServer`. No behavior change without options; existing call sites are untouched. The motivation is letting a downstream daemon embed the metering logic as a library with a database-backed store and a dynamically refreshed price table, rather than forking the binary.

Roasbeef:
Pushed two fixes for the CI failures on this branch.

The RPC check job was failing because the committed pricesrpc/prices.pb.go and prices.pb.gw.go had drifted from the protoc-gen-go and grpc-gateway versions pinned in go.mod. Regenerating locally through pricesrpc/gen_protos_docker.sh (same dockerized toolchain CI uses) produced a diff that matches CI's own reported diff line for line. Committed the regenerated stubs as is, no hand edits.

The unit-race failures (TestCheckMeteredAccess, TestReportUsageWithRetry, TestUsageObservingBody) were a real data race. usageObservingBody.report spawns a detached goroutine (reportUsageWithRetry) that read the package-level reportInitialBackoff variable, while TestReportUsageWithRetry mutated that same variable to shrink its backoff. Since the report goroutine outlives the request and the test that triggered it, a goroutine left over from an earlier test could still be reading the global while the next test wrote to it. Fixed by replacing the two package-level knobs with a reportSchedule struct that's passed explicitly into reportUsageWithRetry, defaulting to defaultReportSchedule in production. Tests now build their own schedule and pass it in rather than mutating shared state.

Verified proxy, meterd, and pricer with -race locally (bounded runs, not the full proxy suite since some of those tests are known to fail locally for environmental reasons unrelated to this PR).

Roasbeef:
Four follow-up commits on top, tightening the metering accounting after another review pass:

- 109de2a adds a reserved_estimate field to the authorize response and usage report, so the price server releases the exact reservation an authorization took instead of an approximate default. Without the correlation, mismatched estimates accumulate as residue until a paid bundle reads as exhausted with balance left. Zero-valued echoes keep old proxies and pricers interoperating.
- b7da329 weights debits by direction: bundles are priced at the blended rate but the buyer picks the prompt/completion mix, so raw token debits let an output-heavy workload consume more msat value than it pays for. The debit now charges tokens at their directional msat value against the blended rate, rounding up.
- 10394b9 makes model resolution fail closed: a request naming a model the rate source does not know is denied rather than falling back to the default model, which previously let an unpriced upstream model ride a cheap bundle through the mismatch guard.
- a37c051 adds an eviction age floor so a mint-spam burst cannot evict a just-paid bundle in the window before its first authorized request.

PR #257, opened 2026-07-29
challenger: mint L402 invoices from a Wavelength wallet

Roasbeef:
In this PR, we let aperture take its challenge invoices from a Wavelength wallet daemon instead of an lnd node, so a seller can charge for a service without running Lightning infrastructure at all.

The motivation is what the seller no longer has to operate. An invoice minted through walletdk's `Recv` is backed by an inbound swap: the swap server supplies the inbound liquidity and settles the payment into the wallet's Ark balance. The seller therefore needs no Lightning node, no channels, and no inbound capacity, which is usually the most annoying part of standing an L402 seller up.

Configuration is one line in place of three:

```yaml
authenticator:
  network: "regtest"
  wavelengthgateway: "http://localhost:10061"
  # no lndhost, no tlspath, no macdir
```

## How it hangs together

We hang this off the existing `InvoiceClient` seam rather than writing a second `Challenger`, so `LndChallenger` keeps owning the mint and verify logic and the new code is only a transport. `AddInvoice` posts to the wallet's `/v1/wallet/recv` and maps the reply back onto `lnrpc.AddInvoiceResponse`. The wallet returns the payment hash next to the payment request, so we read it from there rather than decoding the bolt11, and the two can never disagree about which hash the challenge commits to.

## On strictverify

`strictverify` is refused at construction rather than ignored. Strict verification means refusing a token whose invoice was never seen to settle, and `LndChallenger` implements that by listing and subscribing to invoices, neither of which this backend can do. Accepting the flag would leave aperture believing it was verifying settlement when it was not, and the symptom would be tokens honoured for invoices nobody paid. `ListInvoices` and `SubscribeInvoices` return errors for the same reason: answering with an empty list would read to the challenger as "no such invoice".

That does mean this backend is currently non-strict only, i.e. it trusts that possession of the preimage implies payment. That is the same assumption the lnd backend makes under the default `strictverify: false`. Wiring settlement lookups through the wallet's entry stream would lift the restriction and is the obvious follow-up.

## Testing

Unit tests cover the happy path, the millisatoshi fallback, a wallet error reaching the operator intact, malformed and short payment hashes, the `strictverify` guard, gateway URL validation, and the unsupported tracking methods.

It has also been run end to end on regtest against a real OpenAI-compatible upstream, driven by an external buyer client, with a metered dynamic pricer minting prepaid token bundles. A cold start minted an 11,500 sat challenge from the seller's wallet, the buyer paid it, and the balances moved exactly (buyer 52,000 to 40,500, seller 0 to 11,500, zero fee) with the receipt's preimage verifying against the payment hash.

The interesting part of that run: both ends were Ark wallets on the same operator, so the payment settled as an in-ark vHTLC carrying the invoice's own payment hash rather than routing over Lightning. The preimage still came back, so L402's proof-of-payment held over a path with no Lightning hop in it at all.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -377,8 +377,22 @@ func (a *Aperture) Start(errChan chan error, shutdown <-chan struct{}) error {
 				"minting invoices from %v",
 				authCfg.WavelengthGateway)
 
+			// An unauthenticated gateway is a deliberate
+			// configuration on the wallet's side, and the daemon
+			// refuses it on mainnet, so say plainly which one this
+			// is rather than leaving the operator to discover it
+			// as a 500 on the first request for a paid resource.
+			if authCfg.WavelengthMacaroonPath == "" {
+				log.Warnf("No wavelengthmacaroonpath set, so " +
+					"invoice requests will carry no " +
+					"credential: this only works against " +
+					"a wallet daemon started with " +
+					"rpc.no-macaroons")
+			}
+
 			client, err := challenger.NewWavelengthInvoiceClient(
 				authCfg.WavelengthGateway, a.cfg.StrictVerify,
+				authCfg.WavelengthMacaroonPath,
 			)
 			if err != nil {
 				return fmt.Errorf("unable to create "+
--- a/aperture.go
+++ b/aperture.go
@@ -372,6 +372,28 @@ func (a *Aperture) Start(errChan chan error, shutdown <-chan struct{}) error {
 					"challenger: %w", err)
 			}
 
+		case authCfg.WavelengthGateway != "":
+			log.Infof("Using wavelength's authenticator config, "+
+				"minting invoices from %v",
+				authCfg.WavelengthGateway)
+
+			client, err := challenger.NewWavelengthInvoiceClient(
+				authCfg.WavelengthGateway, a.cfg.StrictVerify,
+			)
+			if err != nil {
+				return fmt.Errorf("unable to create "+
+					"wavelength invoice client: %w", err)
+			}
+
+			a.challenger, err = challenger.NewLndChallenger(
+				client, a.cfg.InvoiceBatchSize, genInvoiceReq,
+				context.Background, errChan, a.cfg.StrictVerify,
+				challengerOpts...,
+			)
+			if err != nil {
+				return err
+			}
+
 		case authCfg.LndHost != "":
 			log.Infof("Using lnd's authenticator config")
 
--- a/challenger/wavelength.go
+++ b/challenger/wavelength.go
@@ -1,0 +1,225 @@
+package challenger
+
+import (
+	"bytes"
+	"context"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"google.golang.org/grpc"
+)
+
+const (
+	// recvPath is the wallet daemon's HTTP/JSON route for requesting an
+	// inbound payment.
+	recvPath = "/v1/wallet/recv"
+
+	// defaultRecvTimeout bounds a single invoice request. Minting happens
+	// inline while a client waits on its 402, so this is deliberately
+	// short: a wallet that has gone away should surface as a failed
+	// challenge rather than holding the request open.
+	defaultRecvTimeout = 30 * time.Second
+)
+
+// WavelengthInvoiceClient mints L402 challenge invoices from a Wavelength
+// wallet daemon instead of from an lnd node.
+//
+// The point of it is what the seller no longer has to run. An invoice from
+// waved is backed by an inbound swap, so the swap server supplies the inbound
+// liquidity and settles the payment into the wallet's Ark balance. The seller
+// therefore needs no Lightning node, no channels, and no inbound capacity to
+// get paid, which is the part of standing up an L402 seller that usually costs
+// the most effort.
+//
+// It satisfies InvoiceClient, so it drives the existing LndChallenger rather
+// than duplicating it. Only AddInvoice is reachable: the other two methods
+// exist to track invoice state, which LndChallenger only does under strict
+// verification, and this backend refuses that mode outright (see
+// NewWavelengthInvoiceClient).
+type WavelengthInvoiceClient struct {
+	// recvURL is the fully resolved endpoint invoices are requested from.
+	recvURL string
+
+	client *http.Client
+}
+
+// A compile-time check that we satisfy the interface the challenger needs.
+var _ InvoiceClient = (*WavelengthInvoiceClient)(nil)
+
+// NewWavelengthInvoiceClient creates an invoice client against the wallet
+// daemon's HTTP/JSON gateway, given its base URL (for example
+// http://localhost:10061).
+//
+// strictVerify is rejected rather than ignored. Strict verification means the
+// challenger watches every invoice's state and refuses a token whose invoice
+// it has not seen settle, which it does by listing and subscribing to
+// invoices. This backend can do neither, so accepting the flag would leave
+// aperture believing it was verifying settlement when it was not: the failure
+// would be silent and would only show up as tokens honoured for invoices that
+// were never paid.
+func NewWavelengthInvoiceClient(gateway string,
+	strictVerify bool) (*WavelengthInvoiceClient, error) {
+
+	if strictVerify {
+		return nil, fmt.Errorf("strictverify is not supported with " +
+			"the wavelength authenticator: the wallet daemon " +
+			"exposes no invoice subscription to verify " +
+			"settlement against")
+	}
+
+	if gateway == "" {
+		return nil, fmt.Errorf("wavelength gateway URL is required")
+	}
+
+	parsed, err := url.Parse(gateway)
+	if err != nil {
+		return nil, fmt.Errorf("invalid wavelength gateway URL %q: %w",
+			gateway, err)
+	}
+	if parsed.Scheme != "http" && parsed.Scheme != "https" {
+		return nil, fmt.Errorf("wavelength gateway URL %q must be "+
+			"http or https", gateway)
+	}
+	if parsed.Host == "" {
+		return nil, fmt.Errorf("wavelength gateway URL %q has no host",
+			gateway)
+	}
+
+	return &WavelengthInvoiceClient{
+		recvURL: strings.TrimRight(parsed.String(), "/") + recvPath,
+		client:  &http.Client{Timeout: defaultRecvTimeout},
+	}, nil
+}
+
+// recvRequest asks the wallet for an inbound payment of a given size.
+type recvRequest struct {
+	AmountSat int64  `json:"amt_sat"`
+	Memo      string `json:"memo,omitempty"`
+}
+
+// recvResponse is the wallet's reply. The payment hash is nested under the
+// ledger entry that tracks the receive, and is read from there rather than
+// decoded out of the payment request, so the two can never disagree about
+// which hash this challenge commits to.
+type recvResponse struct {
+	Invoice string `json:"invoice"`
+	Entry   struct {
+		Request struct {
+			LightningInvoice struct {
+				PaymentHash string `json:"payment_hash"`
+			} `json:"lightning_invoice"`
+		} `json:"request"`
+	} `json:"entry"`
+}
+
+// gatewayError is the error shape the gateway returns for a failed call.
+type gatewayError struct {
+	Message string `json:"message"`
+}
+
+// AddInvoice mints a challenge invoice, translating lnd's request and response
+// shapes onto the wallet's receive call.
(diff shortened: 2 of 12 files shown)


Roasbeef:
Good catch, and it reproduces exactly as described. Fixed in e685963.

I confirmed the failure path first: with the settlement callback set,
`Start()` skips its early return (which needs both `!strictVerify` and a nil
`onSettled`), falls into the reconciliation loop, and `ListInvoices` returns
`listing invoices is not supported by the wavelength authenticator`, which
propagates out of `NewLndChallenger` and takes the process down. The two new
tests fail with precisely that string when the fix is reverted.

I went with a variant of the second suggestion rather than filtering the opts
at the call site. Stripping the option in `aperture.go` works, but it leaves
every future caller to remember which backend it wired up, and the next
invoice client that mints somewhere it cannot enumerate walks into the same
trap. So the capability is now declared by the client:

```go
// InvoiceTracker is an optional capability of an InvoiceClient. A client that
// mints its invoices somewhere it cannot then enumerate them, such as a wallet
// daemon reached over an HTTP gateway, implements it and reports false.
type InvoiceTracker interface {
	TracksInvoices() bool
}
```

It is a separate interface rather than a method on `InvoiceClient` because the
lnd-backed clients are `lnrpc` types we do not own. A client that says nothing
is taken to track its invoices, so every existing path behaves exactly as
before.

`NewLndChallenger` then resolves it once, before the settlement queue is
started, so `Start()`'s existing early return does the rest:

- Strict verification with a non-tracking client is now a hard error rather
  than a silent downgrade. Skipping the tracking while leaving the flag on
  would have aperture believing it verified settlement while honouring tokens
  for invoices nobody paid.
- A settlement callback is dropped with a warning. The dashboard losing one
  column beats the proxy refusing to start, and there is genuinely nothing to
  record: the gateway exposes minting and nothing else.

Three tests cover it: the admin-dashboard combination (constructor succeeds,
neither `ListInvoices` nor `SubscribeInvoices` is attempted, no settlement
queue is left running, minting still works), the strict-verify refusal, and a
regression test that an ordinary tracking client still reconciles its
historical settled invoices through the constructor.

aperture.go
```diff
@@ -357,6 +357,28 @@ func (a *Aperture) Start(errChan chan error, shutdown <-chan struct{}) error {
...
+
+			client, err := challenger.NewWavelengthInvoiceClient(
+				authCfg.WavelengthGateway, a.cfg.StrictVerify,
+			)
+			if err != nil {
+				return fmt.Errorf("unable to create "+
+					"wavelength invoice client: %w", err)
+			}
+
+			a.challenger, err = challenger.NewLndChallenger(
```
litbot-9000, line 402: **Enabling the admin API breaks startup on this backend.**

`buildChallengerOpts` (aperture.go:950) appends `WithSettlementCallback` whenever `admin.enabled` is true and `txnStore != nil`, and those opts are passed straight through here. That makes `challenger.onSettled != nil`, and `LndChallenger.Start()` only skips invoice tracking when `!strictVerify && onSettled == nil` (challenger/lnd.go:257). So it enters the reconciliation loop and calls `l.client.ListInvoices`, which `WavelengthInvoiceClient` always fails:

```
unable to start challenger: listing invoices is not supported by the wavelength authenticator
```

`Aperture.Start` returns that error and the daemon never comes up. Any operator running `wavelengthgateway` together with the admin dashboard hits this on first boot.

The doc comment on `WavelengthInvoiceClient` (challenger/wavelength.go:41) says the tracking methods are unreachable because `LndChallenger` only tracks state under strict verification — that is only true with the admin API off. Either refuse the combination explicitly in config validation, or skip the settlement callback for this backend.

challenger/wavelength.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, fmt.Errorf("marshal recv request: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(
+		ctx, http.MethodPost, w.recvURL, bytes.NewReader(body),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("build recv request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
```
litbot-9000, line 217: **No macaroon is ever sent, so this fails against a default waved.**

waved enforces macaroon auth on its RPC unless `rpc.no-macaroons` is set (`waved/rpc_security.go:49`), and `rpc.no-macaroons` is rejected outright for mainnet TCP listeners unless `allow-insecure-mainnet` is set (`waved/config.go:1656`). The HTTP/JSON gateway does forward a `macaroon` header into gRPC metadata (`waved/gateway_server.go:167`), but this client only ever sets `Content-Type`, and `AuthConfig` exposes no knob to supply one.

Concretely: point aperture at a stock `waved` and the first request to a paid resource produces `wallet recv failed (401): ...` from `AddInvoice`, `NewChallenge` fails, and the client gets a 500 instead of a 402. The integration only works against a waved deliberately started with macaroons disabled, which is not permitted on mainnet.

Suggest a `wavelengthmacaroonpath` (or hex) option, read at startup and sent as `req.Header.Set("macaroon", hex)`.

challenger/wavelength.go
```diff
@@ -0,0 +1,225 @@
...
+func (w *WavelengthInvoiceClient) AddInvoice(ctx context.Context,
+	in *lnrpc.Invoice, _ ...grpc.CallOption) (*lnrpc.AddInvoiceResponse,
+	error) {
+
+	// The mint prices challenges in satoshis, but tolerate a millisatoshi
+	// amount rather than silently minting a zero-value invoice, which the
+	// wallet would reject with a far less obvious error.
+	amountSat := in.Value
+	if amountSat == 0 && in.ValueMsat > 0 {
+		amountSat = in.ValueMsat / 1000
```
litbot-9000: **Truncating division undercharges.**

`in.ValueMsat / 1000` rounds toward zero. 1500 msat mints a 1 sat invoice (a 33% undercharge), and anything under 1000 msat lands on `amountSat <= 0` and fails the challenge with `invoice amount must be positive, got 0 sat` despite a positive price having been requested — exactly the "silently mint a zero-value invoice" outcome the comment above says it is guarding against.

```suggestion
		amountSat = (in.ValueMsat + 999) / 1000
```

proxy/proxy.go
```diff
@@ -652,16 +652,29 @@ func matchService(req *http.Request, services []*Service) (*Service, bool) {
...
+// wins.
 func addCorsHeaders(header http.Header) {
 	log.Debugf("Adding CORS headers to response.")
 
-	header.Add("Access-Control-Allow-Origin", "*")
-	header.Add("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
-	header.Add("Access-Control-Expose-Headers",
+	header.Set("Access-Control-Allow-Origin", "*")
+	header.Set("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
+	header.Set("Access-Control-Expose-Headers",
```
litbot-9000: **`Set` is right for `Allow-Origin`, but it drops backend values on the list-valued fields.**

`Access-Control-Allow-Origin` is single-valued, so replacing it is correct. `Access-Control-Expose-Headers` (and `Allow-Headers`/`Allow-Methods`) are list-based fields: per RFC 7230 and the Fetch spec's "get, decode, and split", repeated field lines are combined with `", "`, so duplicates there were *not* invalid — they merged.

So a backend that responds with `Access-Control-Expose-Headers: X-Request-Id` used to reach the browser as `X-Request-Id, WWW-Authenticate, Payment-Receipt` and JS could do `resp.headers.get('X-Request-Id')`. After this change aperture overwrites the field with only its own list and that read returns `null` — a regression for every backend that exposes its own headers.

Worth appending to the existing value on the list-valued fields rather than replacing:

```go
appendCSV(header, "Access-Control-Expose-Headers", "WWW-Authenticate, Payment-Receipt")
```

The new `TestProxyCORSNotDuplicated` would not catch this, since the stub backend never sets `Expose-Headers`.

challenger/wavelength_test.go
```diff
@@ -0,0 +1,222 @@
...
+// it, along with a pointer to the last request body it received.
+func newTestWallet(t *testing.T, handler http.HandlerFunc) (
+	*WavelengthInvoiceClient, *string) {
+
+	t.Helper()
+
+	var lastBody string
+	server := httptest.NewServer(http.HandlerFunc(
+		func(w http.ResponseWriter, r *http.Request) {
+			buf := make([]byte, r.ContentLength)
```
litbot-9000: A single `Read` is not guaranteed to fill `buf`, so `lastBody` can be a truncated prefix and the `require.Contains` assertions on it can fail intermittently; they pass today only because the bodies are a few dozen bytes. And if any future case sends a chunked/unknown-length body, `r.ContentLength` is `-1` and `make([]byte, -1)` panics in the handler goroutine.

```suggestion
			buf, _ := io.ReadAll(r.Body)
			lastBody = string(buf)
```
(needs `io` in the import block)

Roasbeef:
Thanks, this was a good review. All five reproduce, and all five are fixed.
Pushed as four commits on top of `e685963`:

| Finding | Commit |
| --- | --- |
| Admin dashboard + Wavelength crashes at startup | [`e685963`](https://github.com/lightninglabs/aperture/commit/e685963) (pushed before this review) |
| Truncating division undercharges | [`dbf341f`](https://github.com/lightninglabs/aperture/commit/dbf341f) |
| No macaroon sent to the gateway | [`b0de0d6`](https://github.com/lightninglabs/aperture/commit/b0de0d6) |
| `Set` drops backend values on list-valued CORS fields | [`1332663`](https://github.com/lightninglabs/aperture/commit/1332663) |
| Single `Read` can truncate the test body | [`235b0e4`](https://github.com/lightninglabs/aperture/commit/235b0e4) |

## The CORS one was my error, and the correction is the right one

You are right and my commit message was wrong. I reasoned from
`Access-Control-Allow-Origin`, which is single-valued, and then applied
`Set` to three list-based fields where duplicates had always merged rather
than been invalid. So the change fixed one field and broke three.

`Expose-Headers` is the one that stings, exactly as you describe: only the
backend knows that list, and dropping its entries makes those headers
unreadable from JS with no error anywhere. `headers.get()` returns `null`,
so it reads as the backend having stopped sending the header.

`mergeCorsList` now keeps the backend's entries, appends ours, and skips any
it had already named, matching case-insensitively since header names are.
The backend's entries stay at the front so the list reads in the order that
service published it, and the result goes out as one field line.

You were also right that `TestProxyCORSNotDuplicated` could not catch it: the
stub backend never set `Expose-Headers`. It now sets
`Access-Control-Expose-Headers: X-Request-Id`, a header nothing in aperture
could guess belongs on the list, and two tests cover the merge. Reverting the
merge fails them with:

```
[]string{"WWW-Authenticate", "Payment-Receipt"} does not contain "X-Request-Id"
```

## Rounding

Took the ceiling. Rounding up bills at most one satoshi over the quoted
price, the smallest amount the wallet can express, and never under. Overflow
on the addition lands negative and is caught by the existing positive-amount
check, so it still fails closed.

Table test covers 2000 → 2, 1500 → 2, 1001 → 2, 999 → 1, 1 → 1. With the old
truncation the last two fail with `invoice amount must be positive, got 0
sat`, which is the outage you described rather than an undercharge.

## Macaroon

Confirmed against the daemon source before writing anything:
`rpc_security.go` guards macaroon setup on `!NoMacaroons`, `config.go`
refuses `no-macaroons` on a mainnet TCP listener without
`allow-insecure-mainnet`, and `rpcauth.MacaroonMetadataKey` is `"macaroon"`
with `HexFromFile` establishing the encoding.

Added `wavelengthmacaroonpath`. Two details beyond wiring the header:

- The file is read at construction, so a bad path fails at start-up rather
  than on the first 402 a real client is waiting on.
- It is unmarshalled, not just read. The mistake worth catching is naming
  the wrong file, and a TLS certificate hex encodes perfectly well; without
  the parse it comes back from the daemon much later as a 401 that names
  nothing.

Empty still sends no credential, since that is the only way to reach a
daemon with macaroons disabled and is how the regtest and signet setups run
today. Aperture logs a warning naming the consequence at start-up, because
from inside the proxy an intentionally open gateway and an operator who
forgot look identical. I did not make it a hard failure on mainnet: an
operator running `allow-insecure-mainnet` is doing something unwise but not
something aperture can rule out from here.

## Test body read

Switched to `io.ReadAll`. Worth noting the panic you flagged is the sharper
half: `ContentLength` is `-1` for a chunked body and `make([]byte, -1)`
panics in the handler goroutine, where it surfaces as a failed request
rather than as a pointer at that line.

Full `challenger`, `auth`, and `mpp` suites pass. The rounding and CORS
tests were each verified to fail with their fix reverted.

Roasbeef:
All five of the review findings above are addressed on master; the comments anchor to intermediate commits from before the final revision. Invoice tracking is skipped for clients that cannot enumerate (5fcd63d), the msat conversion rounds up (2233851), the gateway gets a macaroon (494d9c0), the list-valued CORS fields merge backend values via mergeCorsList rather than replacing them, and the test handler reads bodies with io.ReadAll.

PR #270, opened 2026-08-13
proxy+auth+meterd: meter MPP charges and carry streamed responses

Roasbeef:
In this PR, we make streamed inference responses first-class through the whole metered pipeline. The SSE byte path itself was already right, the proxy flushes per write and meterd parses the final usage chunk of a stream, but three things around it were not, and each one surfaced while running a live MPP buyer against a metered signet deployment.

## Payment (MPP) charge requests were never metered

The metering pipeline joins mint-time bookings to request-time draw-downs through the L402 macaroon's token ID, an identity only the L402 door can present. The Payment offer in the same 402 mints its own invoice, so an MPP buyer's bundle was booked under a key its credential could never produce: the bundle sat untouched at full balance while the buyer consumed the service, request after request, unmetered. The join key for a charge is now the invoice's payment hash, booked from the challenge header at mint time and derived from the credential's preimage at request time, so the ID is backed by the proof of payment itself.

This composes with a deliberate policy change: on metered services, charge credentials are re-presentable rather than single use. A metered payment buys a bundle whose draw-down is the spend, so consuming the credential on first use would strand everything past one request; the pricer refusing an exhausted bundle is what ends the credential's useful life, exactly as it does for an L402 token on the same service. The strict single-use rule stands everywhere else.

## The write timeout killed healthy streams

`http.Server.WriteTimeout` is one absolute deadline armed as the response begins, so any generation outliving it (30s in the shipped config) was cut off mid-stream no matter how steadily it was flowing. The proxy now rolls the connection's write deadline forward on each write, refusing the extension once the gap since the last write exceeds the window, so the timeout kills stalls rather than length. The refusal half matters: bytes land in the server's buffer without touching the socket, so the deadline is only enforced by the flush after a write, and a post-stall write that could re-arm the deadline would never be caught. The new streaming tests pin all three properties against a real server: incremental delivery, survival past the absolute deadline, and death on stall.

## An abandoned stream was free inference

The usage object arrives in a stream's final chunk, so a client that disconnects mid-generation discards it; the report then found nothing to bill, released the reservation, and debited zero while the seller ate the upstream cost. An incomplete successful response whose tail carries bytes but no usage object now settles at the reservation estimate, which is the ceiling the request itself stated via max_tokens, matching what the session path already did. No bytes, or a failed response, still costs nothing.

The two pre-existing test failures on master (`TestProxyMPPChargeEndToEnd`, `lnc`) reproduce identically on this branch's base and are untouched here.

See each commit message for a detailed description w.r.t the incremental changes.

### DIFF
--- a/aperture.go
+++ b/aperture.go
@@ -1726,10 +1726,9 @@ func createProxy(cfg *Config, services []*proxy.Service,
 		}
 		if len(meteredServices) > 0 {
 			mppAuth.SetReusableChargePolicy(
-				func(serviceName string) bool {
-					_, ok := meteredServices[serviceName]
-					return ok
-				},
+				reusableChargePolicyForServices(
+					meteredServices,
+				),
 			)
 		}
 
@@ -2017,4 +2016,30 @@ func allowCORS(handler http.Handler, origins []string) http.Handler {
 		// chain of handlers.
 		handler.ServeHTTP(w, r)
 	})
+}
+
+// reusableChargePolicyForServices builds the predicate deciding which
+// resource names may re-present a charge credential.
+//
+// The authenticator is handed the resource name, which for a dynamically
+// priced service is the service name with the request path appended
+// (Service.ResourceName), so a metered service is recognized by prefix
+// rather than equality: "inference" covers "inference/v1/chat/completions".
+// The separator is required so a service named "inference" does not
+// accidentally cover one named "inference2".
+func reusableChargePolicyForServices(
+	metered map[string]struct{}) func(string) bool {
+
+	return func(resourceName string) bool {
+		for name := range metered {
+			if resourceName == name {
+				return true
+			}
+			if strings.HasPrefix(resourceName, name+"/") {
+				return true
+			}
+		}
+
+		return false
+	}
 }
--- a/aperture.go
+++ b/aperture.go
@@ -1706,6 +1706,23 @@ func createProxy(cfg *Config, services []*proxy.Service,
 			return nil, nil, err
 		}
 
+		// On metered services the charge credential is the key to a
+		// prepaid usage bundle, so it stays presentable until the
+		// pricer refuses it, exactly like an L402 token. Everywhere
+		// else the spec's strict single-use rule stands.
+		meteredServices := make(map[string]struct{})
+		for _, svc := range services {
+			if svc.DynamicPrice.Enabled && svc.DynamicPrice.Metered {
+				meteredServices[svc.Name] = struct{}{}
+			}
+		}
+		if len(meteredServices) > 0 {
+			mppAuth.SetReusableChargePolicy(
+				func(serviceName string) bool {
+					_, ok := meteredServices[serviceName]
+					return ok
+				},
+			)
 		// The authenticator runs a background sweep over the records of
 		// spent payments, which the caller stops along with the rest of
 		// the proxy.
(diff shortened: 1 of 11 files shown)


Roasbeef:
Live proof on signet, against the seller wallet stack and Baseten upstream, with this branch's binary running under a deliberately short `writetimeout: 5s`:

**MPP charge metering attribution.** A fresh 2,000 sat payment through the Payment door booked its bundle under both identities at mint, and the paid request debited the payment-hash bundle its own preimage derives:

```
22:34:23.951 Booked bundle of 1000000 tokens ... (token c1c59c95...)   <- L402 macaroon ID
22:34:23.952 Booked bundle of 1000000 tokens ... (token e41b10fd...)   <- MPP payment hash
22:34:32.972 Debited 49 tokens (1 sats) from token e41b10fd..., 999951 tokens remaining
```

The receipt's `reference` and the debited token ID are the same hash. Before this branch, the same flow left the MPP bundle untouched at full balance forever.

**Reusable charge on the metered service.** Re-presenting the same credential was served rather than refused as a replay, and each request drew the bundle down. The wiring fix in the last commit came from this run: the first replay was refused because the policy matched bare service names while the authenticator passes resource names.

**Rolling write deadline.** A 3,000-token generation streamed 1,037 SSE chunks over 16.78 seconds through the 5 second write timeout, first chunk at 0.26s, `[DONE]` at 16.78s, then settled:

```
22:37:56.596 Debited 3851 tokens (8 sats) from token e41b10fd..., 995768 tokens remaining
```

More than three times the absolute deadline; the unmodified server kills this connection at 5s.

**Abandoned stream.** A stream killed 3 seconds into a long generation settled as an incomplete response rather than for free:

```
22:37:14.384 Usage report for token e41b10fd... covers an incomplete response (status 200)
22:37:14.384 Debited 10 tokens (1 sats) from token e41b10fd...
```

This upstream reports running usage totals per chunk, so the partial usage was billed exactly; the reservation-estimate settle added in this branch is the fallback for upstreams that only attach usage to the final chunk, and is pinned by the meterd tests.

Roasbeef:
The adversarial pass on this branch caught a real one: the restructured credential resolution gated the L402 parse on the `Authorization` scheme, but `l402.FromHeader` also reads a hex macaroon from the `Macaroon` and `Grpc-Metadata-Macaroon` headers, so a paid token moved one header over authenticated and then walked past metering. Fixed in the latest commit: the L402 parse runs unconditionally, the presence test covers all three forms, and the new tests pin the hex-plus-preimage-caveat encoding the alternate headers actually carry, which is exactly why nothing had caught it before. The same ordering also means a request that authenticated through the Payment door is metered under it even with a stray unparseable L402 header riding along.

On the two design questions the review asked for an author answer on:

**Charge credentials are not service-bound.** Intended for now, with eyes open. The exposure is bounded the way the review says: one balance, model-mismatch blocks cross-model use, so a bundle spent from service B is the same sats the buyer already paid, attributed to a different door of the same seller. L402's service binding lives in the macaroon caveat, which the Payment scheme has no equivalent of today; the natural place to add one is the challenge `opaque` parameter, which is HMAC-bound and echoed verbatim. That lands as its own change rather than riding this branch.

**Abandonment settles at the full reservation.** Intended, and deliberately seller-favoring: the alternative directions are billing zero (the bug this fixes) or trusting a partial count the buyer controls by choosing when to disconnect. The buyer names the ceiling with max_tokens, so the penalty is self-chosen, and upstreams that report running usage per chunk (Baseten does) never hit the estimate path at all, since the tail then carries a parseable usage object and the exact partial count is billed. The operator note on sizing writetimeout to worst-case inter-chunk gap is in; an HTTP/2 streaming test is queued as a follow-up alongside the ephemeral-port fix for the pre-existing proxy test collisions.

PR #285, opened 2026-09-20
mint: fix rare (1/256) flake in TestTamperedL402

s1ns3nz0:
## Summary

`TestTamperedL402` tampers a minted macaroon's signature to assert
`VerifyL402` rejects it, by overwriting the last serialized byte with a
fixed `0x00`:

```go
macBytes[len(macBytes)-1] = 0x00
```

The signature is HMAC output — effectively random — so there's a 1/256
chance its last byte already *was* `0x00` before the overwrite. When that
happens the "tamper" is a no-op, the macaroon verifies successfully,
`VerifyL402` returns a `nil` error, and the very next line:

```go
if !strings.Contains(err.Error(), "signature mismatch") {
```

panics with a nil pointer dereference instead of failing the test normally.

## Reproduction

Confirmed on an unmodified checkout:

```
go test ./mint/... -run TestTamperedL402 -count=2000
```

fails with a SIGSEGV panic inside `err.Error()`, roughly matching the
expected ~1-in-256-per-run rate.

## Fix

XOR the byte instead of overwriting it, so it's guaranteed to change
regardless of its original value:

```go
macBytes[len(macBytes)-1] ^= 0xFF
```

## Testing

```
go test ./mint/... -run TestTamperedL402 -count=3000   # clean
go test ./mint/... -race -count=10                      # clean
```

Single-line test-only fix, no production code touched.


Roasbeef:
Approved the CI run!

